PR #1, opened 2024-11-20
Restricted commits on main

KOVALW:
## Overview
We don't want to allow accidental changes to main and therefore want to prevent commits.

## Changes
Added `no-commit-to-branch` line to the pre-commit repo calls.


ChiragKumar9:
Closing because this is taken care of in rulesets.

PR #3, opened 2024-11-21
Add parameter loading from command-line specified input file

ChiragKumar9:


KOVALW (commented):
This does look good to me but I'm getting an error while building that seems to be out of scope for the project (`file_create_new` in the ixa source code). Is this going to be problematic?

```
Compiling ixa v0.0.1 (https://github.com/cdcgov/ixa#1ff09c58)
error[E0658]: use of unstable library feature 'file_create_new'
   --> /home/AD71/.cargo/git/checkouts/ixa-58998f6879320292/1ff09c5/src/report.rs:131:36
    |
131 |         let file_creation_result = File::create_new(&path);
    |                                    ^^^^^^^^^^^^^^^^
    |
    = note: see issue #105135 <https://github.com/rust-lang/rust/issues/105135> for more information
```

KOVALW:
I just ran the suggested input from the readme `cargo run -- -i ./input/input.json -o ./output -p incidence`. The `main` should only be initializing the parameters and then shutting down, so I'm confused why that would call in this error.

src/main.rs
```diff
@@ -1,15 +1,52 @@
...
+    #[arg(short, long)]
+    input_file: PathBuf,
+
+    /// path to the output directory
+    #[arg(short, long)]
+    output_directory: PathBuf,
+
+    /// name of the person property periodic report file
+    #[arg(short, long)]
+    person_report: String,
```
confunguido, line 22: Let's define this in the input.json file

confunguido (changes requested):
lgtm after removing -p

KOVALW (approved):
LGTM - thanks for the help with switching versions!

confunguido (approved):

PR #5, opened 2024-11-21
Population loader skeleton

KOVALW:
## Overview
Takes in a csv input person record and creates people in the context. Needs to be rebased to not include changes already accepted in PR #3 

## Changes
Made the `population_loader.rs`, which initializes by reading in `Parameters`, establishing and reading from the record path in `record_dir`, then for each `result`, a new person is created with properties `Age`, `HomeId`, and a derived property `CensusTract`. Each person is by default `Alive=true`.

The parameters loader now requires listing the synthetic population file location as a string. The JSON input file is changed to include this expected string.

The main calls the `population_loader::init(&mut context)` to initialize the population from the person record.

##Next steps
- Census tract may not need to be a derived property
- Tests required to demonstrate functionality
- We will likely want to have `pub enum` for categories to interface with infection


KOVALW:
Closing to simplify rebase on new branch

PR #6, opened 2024-11-21
Population loader skeleton

KOVALW:
## Overview
Takes in a csv input person record and creates people in the context. This is a re-request of PR #5 to simplify linear history 

## Changes
Made the `population_loader.rs`, which initializes by reading in `Parameters`, establishing and reading from the record path in `record_dir`, then for each `result`, a new person is created with properties `Age`, `HomeId`, and a derived property `CensusTract`. Each person is by default `Alive=true`.

The parameters loader now requires listing the synthetic population file location as a string. The JSON input file is changed to include this expected string.

The main calls the `population_loader::init(&mut context)` to initialize the population from the person record.

## Next steps
- Census tract may not need to be a derived property
- Tests required to demonstrate functionality
- We will likely want to have `pub enum` for categories to interface with infection


confunguido (approved):
I made a few corrections. Think we can merge but should probably start creating a simple report to output demographics from population and a couple of tests to make sure we are reading the population correctly.

src/population_loader.rs
```diff
@@ -0,0 +1,44 @@
...
+    homeId: usize,
+}
+
+define_person_property!(Age, u8);
+define_person_property!(HomeId, usize);
+define_person_property_with_default!(Alive, bool, true);
+
+define_derived_property!(CensusTract, usize, [HomeId], |home_id| { home_id / 10000 });
+
+pub fn create_person_from_record(context: &mut Context, person_record: &PeopleRecord) -> PersonId {
```
ChiragKumar9, line 24: Doesn't need to be a `pub` function.

src/population_loader.rs
```diff
@@ -0,0 +1,44 @@
...
+        .unwrap()
+}
+
+pub fn init(context: &mut Context) {
+    let parameters = context
+        .get_global_property_value(Parameters)
+        .unwrap()
+        .clone();
+
+    let mut reader = csv::Reader::from_path(parameters.synth_population_file.clone()).unwrap();
```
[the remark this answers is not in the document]
ChiragKumar9: Meaning that the function returns `Ok(())` at the end if nothing has gone wrong and the `.unwrap()` is replaced with a `?`

src/population_loader.rs
```diff
@@ -0,0 +1,44 @@
...
+define_person_property!(Age, u8);
+define_person_property!(HomeId, usize);
+define_person_property_with_default!(Alive, bool, true);
+
+define_derived_property!(CensusTract, usize, [HomeId], |home_id| { home_id / 10000 });
+
+pub fn create_person_from_record(context: &mut Context, person_record: &PeopleRecord) -> PersonId {
+    context
+        .add_person(((Age, person_record.age), (HomeId, person_record.homeId)))
+        .unwrap()
```
ChiragKumar9, line 27: Same comment about `expect` over `unwrap` here. Whether you want this to propagate an error is up to you -- I think it should, especially if we will end up using this function to potentially create people via births in the future (so the function's signature would be `Result<PersonId, IxaError>`). But, I can see an argument where you just want to use `expect` here since we've already validated the data down below in the `init` function.

src/population_loader.rs
```diff
@@ -0,0 +1,44 @@
...
+
+pub fn init(context: &mut Context) {
+    let parameters = context
+        .get_global_property_value(Parameters)
+        .unwrap()
+        .clone();
+
+    let mut reader = csv::Reader::from_path(parameters.synth_population_file.clone()).unwrap();
+
+    for result in reader.deserialize() {
```
ChiragKumar9, line 38: I am sure @confunguido has more thoughts here, but I would be curious to see if using a map iterator is faster! Of course, does not have to be addressed in this PR.

src/population_loader.rs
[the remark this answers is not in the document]
confunguido: I was gonna implement ixaerror yesterday but decided to leave as is for now, but good point.

src/population_loader.rs
confunguido: You're right

src/population_loader.rs
confunguido: agree with ixaerror being the right way to go here

src/population_loader.rs
KOVALW: Created an [issue](https://github.com/CDCgov/ixa/issues/92) in the Ixa repo to cover this but csv errors aren't included yet. Going with unwrap() for now.

.pre-commit-config.yaml
```diff
@@ -66,4 +66,4 @@ repos:
         args: ["--verbose", "--"]
     -   id: clippy
         # Use 'all-targets' to run on all code, including tests and examples
-        args: ["--all-targets", "--", "-D", "warnings", "-W", "clippy::pedantic"]
+        args: ["--all-targets", "--","-A", "non-snake-case", "-D", "warnings", "-W", "clippy::pedantic"]
```
confunguido, line 69: I don't think this is right. We want to keep non-snake-case as a warning and disable only where needed

src/population_loader.rs
```diff
@@ -0,0 +1,69 @@
...
+    fn test_create_person_from_record() {
+        let mut context = Context::new();
+        context.init_random(0);
+        let record = PeopleRecord {
+            age: 42,
+            homeId: 12345,
+        };
+        let person_id = create_person_from_record(&mut context, &record).unwrap();
+        assert_eq!(context.get_person_property(person_id, Age), 42);
+        assert_eq!(context.get_person_property(person_id, HomeId), 12345);
```
confunguido, line 67: Can we check the census tract? I think we may want to put some guardrails here to make sure homeId has the expected format.

confunguido (changes requested):

.pre-commit-config.yaml
KOVALW: I think it's quite difficult to do this in the macro invocation where one of the warnings was arising. Do you have workarounds for calling it with the derived property macro?

src/population_loader.rs
```diff
@@ -0,0 +1,69 @@
...
+    let parameters = context
+        .get_global_property_value(Parameters)
+        .unwrap()
+        .clone();
+
+    let mut reader =
+        csv::Reader::from_path(parameters.synth_population_file).expect("Failed to open file.");
+
+    for result in reader.deserialize() {
+        let record: PeopleRecord = result.expect("Failed to parse record.");
```
ChiragKumar9, line 43:
```suggestion
        let record: PeopleRecord = result?;
```

you should be good here to use `?` because the base error is `Error` not a CSV error, I think.

src/population_loader.rs
ChiragKumar9: I agree

ChiragKumar9 (changes requested):

input/input.json
```diff
@@ -5,5 +5,6 @@
     "r_0": 2.5,
     "infection_duration": 5.0,
     "generation_interval": 5.0,
-    "report_period": 1.0
+    "report_period": 1.0,
+    "synth_population_file": "input/people_test.csv"
```
ChiragKumar9, line 9: `population` should also no longer be required here as an input!

src/population_loader.rs
KOVALW: This does return a csv::Error actually, so I won't be implementing this yet

input/people_test.csv
```diff
@@ -0,0 +1,16 @@
+age,homeId
```
ekr-cfa, line 1: Is this an existing file format? If not:

* Is there a reason to use camelcase here, thus forcing it in serde?
* I would normally think homeId would come first.

Obviously, if this is existing then we have to live with it I guess.

src/main.rs
```diff
@@ -30,6 +32,10 @@ fn initialize(args: &Args) -> Result<Context, IxaError> {
         .clone();
     // model tidyness -- random seed, automatic shutdown
     context.init_random(parameters.seed);
+
+    // load the population from person record in input file
+    population_loader::init(&mut context)?;
```
ekr-cfa, line 37: Perhaps population_loader should be a trait extension on Context?

src/population_loader.rs
```diff
@@ -0,0 +1,140 @@
...
+#[allow(non_snake_case)]
+pub struct PeopleRecord<'a> {
+    age: u8,
+    homeId: &'a [u8],
+}
+
+define_person_property!(Age, u8);
+define_person_property!(HomeId, usize);
+define_person_property_with_default!(Alive, bool, true);
+
```
ekr-cfa, line 23:
```suggestion
```

src/population_loader.rs
```diff
@@ -0,0 +1,140 @@
...
+pub fn init(context: &mut Context) -> Result<(), IxaError> {
+    let parameters = context
+        .get_global_property_value(Parameters)
+        .unwrap()
+        .clone();
+
+    load_synth_population(context, PathBuf::from(parameters.synth_population_file))?;
+    context.index_property(Age);
+    context.index_property(CensusTract);
```
ekr-cfa, line 71: I'm unsure why this index is being created in this fxn. Shouldn't it be in `main`? Which properties we query isn't a property of loading.

src/population_loader.rs
```diff
@@ -0,0 +1,140 @@
...
+        let tract: usize = 36_093_033_102;
+
+        assert_eq!(
+            age,
+            context.get_person_property(context.get_person_id(0), Age)
+        );
+        assert_eq!(
+            tract,
+            context.get_person_property(context.get_person_id(0), CensusTract)
+        );
```
ekr-cfa, line 109: * Shouldn't you check both people?
* It would be good not to depend on being able to convert integers to person ids.
* If you create a `person_id` temporary, you won't need to call `get_person_id()` repeatedly.

src/population_loader.rs
```diff
@@ -0,0 +1,140 @@
...
+            b"
+age,homeId
+43,360930331020001
+42,360930331020002",
+        )
+        .unwrap();
+        let synth_file = path.join("synth_pop_test.csv");
+        load_synth_population(&mut context, synth_file).unwrap();
+        let age: u8 = 43;
+        let tract: usize = 36_093_033_102;
```
ekr-cfa, line 100: Why do you need these temporaries?

src/population_loader.rs
```diff
@@ -0,0 +1,140 @@
...
+            context.get_person_property(context.get_person_id(0), Age)
+        );
+        assert_eq!(
+            tract,
+            context.get_person_property(context.get_person_id(0), CensusTract)
+        );
+    }
+
+    #[test]
+    fn check_invalid_home_id() {
```
ekr-cfa, line 113: What is this testing. From the name I would expect failure, but it succeeds.

ekr-cfa (changes requested):

input/people_test.csv
confunguido: This is the format of the ASPRs synthetic populations. We are ignoring here in this example the rest of the fields. I think we can change that, but that's the reason for now.

src/population_loader.rs
confunguido: * Yes, we should check both people
* Agree
* To clarify, do you suggest we do something like `let person_id = PersonId{0}`  (I don't think that works), or do you suggest something different?

src/population_loader.rs
```diff
@@ -0,0 +1,140 @@
...
+        (HomeId, home_id),
+        (CensusTract, tract),
+    ))?;
+    Ok(person_id)
+}
+
+fn load_synth_population(context: &mut Context, synth_input_file: PathBuf) -> Result<(), IxaError> {
+    let mut reader =
+        csv::Reader::from_path(synth_input_file).expect("Failed to open file. No headers found.");
+    let mut raw_record = csv::ByteRecord::new();
```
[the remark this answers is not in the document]
KOVALW: No reason besides following previous examples in the ixa repo examples, we can change this if it impacts performance

src/population_loader.rs
```diff
@@ -0,0 +1,140 @@
...
+        create_person_from_record(context, &record)?;
+    }
+    Ok(())
+}
+
+pub fn init(context: &mut Context) -> Result<(), IxaError> {
+    let parameters = context
+        .get_global_property_value(Parameters)
+        .unwrap()
+        .clone();
```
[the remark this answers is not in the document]
KOVALW: Is the clone not necessary for keeping context as a mutable borrow when called in `load_synth_population`?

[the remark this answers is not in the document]
ekr-cfa: No, because you are not passing the the reference to parameters to `load_synth_population()` but rather to `PathBuf::new().  This is most easily seen as follows. The current code is:

```
load_synth_population(context, PathBuf::from(parameters.synth_population_file))?;
```

But this is equivalent to:

```
let tmp = PathBuf::from(parameters.synth_population_file);  // immutable borrow here
load_synth_population(context, tmp)?;   // mutable borrow here.
```

input/people_test.csv
ekr-cfa: Fair enough.

src/population_loader.rs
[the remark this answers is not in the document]
ekr-cfa: I don't think it changes performance, it just looks like the code is cleaner.

src/population_loader.rs
ekr-cfa: Whether you want to have an index on a property depends on how you will be querying it, which is a property of the rest of the program, not of the code that loads it, so I think that arguably it should be near the code that is doing the frequent querying.

As an aside, `index_property()` really tells Ixa to index it, but doesn't necessarily cause that index to be computed right away (though I think that's what happens now).

src/population_loader.rs
ekr-cfa: Sorry, I meant why do you need `let age : u8 = 43` rather than just referencing 43 below.

src/population_loader.rs
ekr-cfa: I meant:

```
    let person_id = context.get_person_id(0);

    assert_eq!(
            age,
            context.get_person_property(person_id,  Age)
        );
        assert_eq!(
            tract,
            context.get_person_property(person_id, CensusTract)
        );
```

src/population_loader.rs
KOVALW: For now I've changed these temporaries to be arrays for iterating over both persons.

src/population_loader.rs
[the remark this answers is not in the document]
KOVALW: I think I'm misunderstanding your original suggestion then because simply removing clone and adding a reference `&parameters.synth_population_file` throws the error that context cannot be called as mutable when called as immutable. Should there be an extra temporary? 

`synth_population_file`  is already known to `parameters` as a `PathBuf`, so I'm not sure if `from` would be the first function called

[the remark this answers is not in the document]
ekr-cfa: We've had some version skew because the previous version had `Parameters.synth_population_file` as a string and then had the `PathBuf::from()` inside the call to `load_synth_population`, so what I said won't work.

What will work, however, is to clone just the PathBuf:
https://github.com/CDCgov/ixa-epi-isolation/pull/16

src/population_loader.rs
```diff
@@ -0,0 +1,99 @@
...
+        let (_file, path) = file.keep().unwrap();
+        path
+    }
+
+    #[test]
+    fn check_synth_file_tract() {
+        let mut context = Context::new();
+        let input = String::from("age,homeId\n43,360930331020001\n42,360930331020002");
+        let synth_file = persist_tmp_csv(&input);
+        load_synth_population(&mut context, synth_file).unwrap();
```
ChiragKumar9, line 77: I think you should also check that the population is of size 2.

src/population_loader.rs
```diff
@@ -0,0 +1,99 @@
...
+            assert_eq!(age[id], context.get_person_property(person_id, Age));
+            assert_eq!(
+                tract[id],
+                context.get_person_property(person_id, CensusTract)
+            );
+        }
+    }
+
+    #[test]
+    #[should_panic(expected = "range end index 11 out of range for slice of length 9")]
```
ChiragKumar9, line 92: It's not clear to me what this error message means.

src/population_loader.rs
```diff
@@ -0,0 +1,99 @@
...
+    fn check_invalid_census_tract() {
+        let mut context = Context::new();
+        let input = String::from("age,homeId\n43,360930331\n42,360930331020002");
+        let synth_file = persist_tmp_csv(&input);
+        let _ = load_synth_population(&mut context, synth_file);
```
ChiragKumar9, line 97: Don't you need a `.unwrap()` here at the end of the line? Otherwise, `_` is an error but I don't quite see how it panics in the long-term once you use error propagation.

src/population_loader.rs
ChiragKumar9: Also, I see that this error message isn't being used yet. I assume this is the error message you plan on using when actual error propagation is in place.

src/population_loader.rs
```diff
@@ -0,0 +1,99 @@
...
+use std::path::PathBuf;
+
+#[derive(Deserialize, Debug)]
+#[allow(non_snake_case)]
+pub struct PeopleRecord<'a> {
+    age: u8,
+    homeId: &'a [u8],
+}
+
+define_person_property!(Age, u8);
```
ChiragKumar9, line 17: I wonder if by the same logic that we have indexing the property going in `main`, should we define the property there too? I feel like it is most logical to have the property defined here, but just thinking about best structure. Would be curious what others think.

src/periodic_report_population.rs
```diff
@@ -0,0 +1,89 @@
...
+    let person_census = context.get_person_property(event.person_id, CensusTract);
+    let report_plugin = context.get_data_container_mut(PopulationReportPlugin);
+    report_plugin.census_tract_set.insert(person_census);
+}
+
+pub fn init(context: &mut Context, output_dir: &Path) -> Result<(), IxaError> {
+    let parameters = context
+        .get_global_property_value(Parameters)
+        .unwrap()
+        .clone();
```
ChiragKumar9, line 69: Don't need to clone the whole parameters, just the report period down below.

ChiragKumar9: I didn't get to re-review @confunguido PR, so this should have gone there. But, for now, would you mind making the change here? Thanks!

src/population_loader.rs
k88hudson-cfa: Given that we've designed ixa (including person properties/person property initialization) to support a modular structure, I think it would be better to define properties with the code they're related to v.s. trying to stick everything in main.

ekr-cfa:
@k88hudson-cfa landed the periodic report code in https://github.com/CDCgov/ixa/pull/108, so you should adapt this PR to use that.

src/population_loader.rs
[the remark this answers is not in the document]
KOVALW: We've run into an error that may be the reason for using this variant of reader: that we have a slice with specified lifetime in the PeopleRecord enum. This is not generalized enough for deserialize, and the byte record method gets around this issue. If you have a more elegant solution or think this is fine as is, happy to hear!

src/main.rs
KOVALW: We may want to keep this separate module as a safety feature, to keep anyone from accessing the public extension outside of main and initializing the population from another crate

KOVALW:
> @k88hudson-cfa landed the periodic report code in [CDCgov/ixa#108](https://github.com/CDCgov/ixa/pull/108), so you should adapt this PR to use that.

Since the report branch was merged into this branch already, maybe we should write `update report` as an issue to address after merging in the population loader that is already laden with other changes?

ChiragKumar9 (approved):
Thanks for making all the changes!

src/main.rs
```diff
@@ -18,8 +20,14 @@ struct Args {
     /// path to the output directory
     #[arg(short, long)]
     output_directory: PathBuf,
+
+    #[arg(short, long, default_value = "false", default_missing_value = "true")]
+    force_overwrite: bool,
 }
 
+mod periodic_report_population;
+mod population_loader;
```
ekr-cfa, line 29: Let's put these up at the top of the file before the use directives.

src/main.rs
ekr-cfa: I don't have a strong opinion on if it should be a trait extension (@k88hudson-cfa , wdyt), but I think you can just do `pub(crate)` for the trait, no?

src/parameters.rs
```diff
@@ -1,16 +1,73 @@
...
+    }
+    if parameters.generation_interval <= 0.0 {
+        return Err(IxaError::IxaError(
+            "The generation interval must be positive.".to_string(),
+        ));
+    }
+    Ok(())
+}
+
+define_global_property!(Parameters, ParametersValues, |p| { validate_inputs(p) });
```
ekr-cfa, line 33:
```suggestion
define_global_property!(Parameters, ParametersValues, validate_inputs );
```

You don't need a closure here because `validate_inputs()` already has the right signature.

src/parameters.rs
```diff
@@ -1,16 +1,73 @@
...
+
+#[cfg(test)]
+mod test {
+    use super::validate_inputs;
+    use std::path::PathBuf;
+
+    use crate::parameters::ParametersValues;
+
+    #[test]
+    #[should_panic(expected = "r_0 must be a non-negative number.")]
```
ekr-cfa, line 43: Instead of calling `.unwrap()` and having these panic, I would instead check that they have the right error in `Result<T>`

src/periodic_report_population.rs
```diff
@@ -0,0 +1,93 @@
...
+    global_properties::ContextGlobalPropertiesExt,
+    people::{ContextPeopleExt, PersonCreatedEvent},
+    report::{ContextReportExt, Report},
+};
+use serde::{Deserialize, Serialize};
+use std::path::{Path, PathBuf};
+
+use std::collections::HashSet;
+
+#[derive(Serialize, Deserialize, Clone, PartialEq)]
```
ekr-cfa, line 17: Do you need to implement all of these? It looks like maybe not Deserialize, Clone, or PartialEq

src/periodic_report_population.rs
ekr-cfa, line 1: This function should use @k88hudson-cfa's new periodic report facility.

src/population_loader.rs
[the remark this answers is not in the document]
ekr-cfa: OK.

src/population_loader.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    Ok(())
+}
+
+fn load_synth_population(context: &mut Context, synth_input_file: PathBuf) -> Result<(), IxaError> {
+    let mut reader = csv::Reader::from_path(synth_input_file)?;
+    let mut raw_record = csv::ByteRecord::new();
+    let headers = reader.byte_headers()?.clone();
+
+    //must be re-written for serde without over-generalizing
```
ekr-cfa, line 43: I don't understand what this comment means,

src/population_loader.rs
```diff
@@ -0,0 +1,102 @@
...
+        create_person_from_record(context, &record)?;
+    }
+    Ok(())
+}
+
+pub fn init(context: &mut Context) -> Result<(), IxaError> {
+    let parameters = context.get_global_property_value(Parameters).unwrap();
+
+    load_synth_population(context, parameters.synth_population_file.clone())?;
+    Ok(())
```
ekr-cfa, line 55:
```suggestion
    load_synth_population(context, parameters.synth_population_file.clone())
```

src/population_loader.rs
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    #[test]
+    fn check_synth_file_tract() {
+        let mut context = Context::new();
+        let input = String::from("age,homeId\n43,360930331020001\n42,360930331020002");
+        let synth_file = persist_tmp_csv(&input);
+        load_synth_population(&mut context, synth_file).unwrap();
+        let age = [43, 42];
+        let tract = [36_093_033_102, 36_093_033_102];
```
ekr-cfa, line 80: Shouldn't you also check home_id?

ekr-cfa:
> Can't respond in-line, but re [#6 (comment)](https://github.com/CDCgov/ixa-epi-isolation/pull/6#discussion_r1866160397) @ekr-cfa, if I understand your suggestion correctly, if we make the population loader a trait extension so that we call `context.initialize_pop_loader()` instead of `population_loader::init(&mut context)`, that means we could call the initialize method in _any_ module, right? Comparatively, the current pattern only ever lets us call initialize functions in `main`. I feel that this is safer/less confusing to the user. But, I may be missing something. Would welcome thoughts here.

When you say any module, do you mean any module in this crate?

src/periodic_report_population.rs
KOVALW: Just to be clear, it's best to address issue #22 here then?

ChiragKumar9:
> When you say any module, do you mean any module in this crate?

Yes. I think the trait extension could be called by any module in the crate whereas the current pattern can only be called in main. Or, I may be misunderstanding how to not make trait extensions only useable to main.

src/periodic_report_population.rs
ekr-cfa: Either that or just remove this code, land the remainder of the PR, and then write new code with #22. That might be better in that that would do something and reduce branch lifetime.

ChiragKumar9:
> You can’t do “mod foo” in another module? I actually don’t know

If you do `mod foo` in `bar.rs`, `foo.rs` has to be in path `src/bar/foo.rs`.

src/periodic_report_population.rs
KOVALW: Ok, I've just added in `add_periodic_report` and cut reference to the old periodic_report_populaiton.rs script.

ChiragKumar9 (changes requested):
Please use the command line argument from `force_overwrite` in the `context.report_options().overwrite()` method rather than feeding in a blanket `true`.

ChiragKumar9 (approved):

ekr-cfa (approved):
LGTM. You will probably want to merge PR #23, because I am going to remove `get_person_id()` in https://github.com/CDCgov/ixa/pull/119

PR #7, opened 2024-11-22
Transmission Readme

ChiragKumar9:
Readme for transmission model

### DIFF
--- a/docs/transmission.md
+++ b/docs/transmission.md
@@ -7,16 +7,20 @@ infectious agent and other susceptible agents. Agents become infectious immediat
 upon successful infection attempt, though they may have a transmission probability of
 0 for some time (mimicking an exposed compartment). Once infectious, agents have a pre-assigned
 number of infection attempts. Infection attempt times are drawn from the generation interval.
-Once there are no more infection attempts remaining for the agent, we argue that they do not
-transition to a recovered compartment but rather back to a susceptible compartment but
-with high immunity.
+Once there are no more infection attempts remaining for the agent, we model the agent as having
+their level of immunity against the disease set to, say, 1 but have them transition back
+to the susceptible compartment. This is to simplify the disease model and make it such that
+immunity levels can be re-calculated when contacts are drawn rather than needing to update
+them on a schedule.
 
 ## Workflow
 
 The transmission model is centered around a key workflow:
 1. Watch for an agent becoming infectious (S --> I transition).
-2. For the infectious agent, draw a number of infection attempts. This can be from the
-$R_i$ distribution or the contact distribution.
+2. For the infectious agent, draw a number of infection attempts. This could be from an $R_i$ distribution
+or the distribution of the number of contacts an individual has based on their network connectivity.
+There is also an option for each infection attempt to have a probability of transmission. This serves to
+scale the contact distribution to a potential $R_i$ distribution.
 3. Draw an infection time for the agent. This consists of taking an ordered draw from the generation
 interval. The math for this is abstracted away under a generic `get_next_infection_time` method.
 4. Draw a susceptible contact from the agent's contact group. Contacts could be drawn uniformly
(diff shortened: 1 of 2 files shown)


docs/transmission.md
```diff
@@ -0,0 +1,89 @@
...
+number of infection attempts. Infection attempt times are drawn from the generation interval.
+Once there are no more infection attempts remaining for the agent, we argue that they do not
+transition to a recovered compartment but rather back to a susceptible compartment but
+with high immunity.
+
+## Workflow
+
+The transmission model is centered around a key workflow:
+1. Watch for an agent becoming infectious (S --> I transition).
+2. For the infectious agent, draw a number of infection attempts. This can be from the
```
confunguido, line 18: What does this mean? Ri vs contact distribution?

docs/transmission.md
```diff
@@ -0,0 +1,89 @@
...
+$R_i$ distribution or the contact distribution.
+3. Draw an infection time for the agent. This consists of taking an ordered draw from the generation
```
confunguido, line 20: Is this the time to start adding math here?

docs/transmission.md
```diff
@@ -0,0 +1,89 @@
...
+interval. The math for this is abstracted away under a generic `get_next_infection_time` method.
+4. Draw a susceptible contact from the agent's contact group. Contacts could be drawn uniformly
+from the contact group or with weight based on person characteristics. The actual
+drawing of a contact can be abstracted into a `get_contact` method. Evaluate whether
```
confunguido, line 24: Wondering what this should look like for our specific example, although we don't need to elaborate on this right now

docs/transmission.md
```diff
@@ -0,0 +1,89 @@
...
+the infection attempt is successful. Success is based on:
+    - a potential probability of transmission in the case where we are drawing the
+    number of contacts from a contact distribution,
+    - whether any interventions are present that may reduce transmission,
+    - the contact's immunity level which is updated when the contact is drawn.
+
+    If the infection attempt is successful, the infectious contact is labeled as infectious.
```
confunguido, line 31: The infectious contact or the susceptible contact?

docs/transmission.md
```diff
@@ -0,0 +1,89 @@
...
+5. Repeat steps #3 and #4, decreasing the number of infection attempts left by one each time
+regardless of whether the attempt is successful and ending when the agent has no zero
+infection attempts remaining.
+6. Re-label the agent as susceptible. The immunity manager will set their latent immunity level
```
confunguido, line 35: This isn't entirely clear to me. Why not recovered? If we are not infecting anyone else why wouldn't we move to the next compartment?

docs/transmission.md
```diff
@@ -0,0 +1,89 @@
...
+at a given time based on their person characteristics to get their realized transmissibility
+for use in the workflow described above, but it does not explicitly track which interventions
+are present at what time or how they impact transmissibility.
+
+- The transmission manager also does not track an individual's contacts. Instead, the transmission
+manager calls a generic method called `get_contact` which is in the `mod contact_manager` to obtain
+the individual's contacts.
+
+- The transmission manager does not track an individual's health status. In fact, it is
+completely independent from health status. Instead, health status is tracked in
```
confunguido, line 52: We can remove this as well. I tihnk it's enough for now to say what transmission does

confunguido (changes requested):

docs/transmission.md
KOVALW: Number of infection attempts could be a draw from a distribution over $R_i$ or the number of infection attempts could be some function of the contact network of the transmitter.

docs/transmission.md
KOVALW: Agreed this should be susceptible

docs/transmission.md
KOVALW: The thinking here was that we would only want the transmission manager to have infectious or susceptible compartments, with susceptible individuals having a susceptibility determined by recent infections, innate immunity, etc, that is altered through the immunity manager. We therefore don't necessarily need a recovered compartment and it simplifies including those post-infectious individuals to be in the contact network of future infectors. The person, now susceptible, wouldn't infect anyone else

docs/transmission.md
```diff
@@ -0,0 +1,89 @@
...
+to the value of immunity based on natural infection to mimic a recovered compartment.
+
+## Relationship to other managers/out of scope for the transmission manager
+
+- The transmission manager does not implement interventions. It does not keep track
+of interventions present or how they impact transmission. Instead, those quantities are
+stored in an intervention plugin. The transmission manager can query an individual's transmissibility
+at a given time based on their person characteristics to get their realized transmissibility
+for use in the workflow described above, but it does not explicitly track which interventions
+are present at what time or how they impact transmissibility.
```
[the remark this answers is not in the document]
KOVALW: I can see how this was getting ahead of ourselves

docs/transmission.md
KOVALW: I would argue that these components are important to include so that we don't build other elements of transmission potential (contact manager) or disease progression into the manager. Ok to decrease detail but I think part of "what the transmission manager does" is specifying what it doesn't do

docs/transmission.md
ChiragKumar9: I say we hold off until at least we have our GI meeting and have consensus from the group?

confunguido (approved):

PR #8, opened 2024-11-22
Validate inputs when reading them in

ChiragKumar9:
Provides a skeleton of checking whether inputs are valid based on our knowledge of them prior to them being used in the model in case they will not be used till later in the model running to try to catch errors early.


src/parameters_loader.rs
```diff
@@ -19,8 +19,66 @@ pub struct ParametersValues {
...
+    if parameters.r_0 < 0.0 {
+        return Err(IxaError::IxaError(
+            "r_0 must be a non-negative number".to_string(),
+        ));
+    }
+    // need to think about a better way of validation
+    // in the long term that makes sense for an arbitrary GI
+    if parameters.generation_interval <= 0.0 {
+        return Err(IxaError::IxaError(
+            "r_0 must be a non-negative number".to_string(),
```
confunguido, line 38: It sounds like we don't know yet how to specify generation interval. Should we remove this for now and only leave `r_0`? Or at least change the error so that it doesn't say `r_0`

src/parameters_loader.rs
```diff
@@ -19,8 +19,66 @@ pub struct ParametersValues {
...
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    use crate::parameters_loader::ParametersValues;
+
+    #[test]
+    fn test_validate_r_0() {
+        let parameters = ParametersValues {
+            population: 1000,
```
confunguido, line 60: We are not using population

confunguido (changes requested):
You may have to make sure that remove paramters not used (already fixed in the other PRs).

src/parameters_loader.rs
ChiragKumar9: Yes, agree -- I requested this to be removed in the population loader PR as that makes the most thematic sense, so I think we need to land that first. I will rebase this to main once we land that.

src/parameters_loader.rs
ChiragKumar9: Oops good catch, thanks! I like having this code there as a template for now and then we can change it as we decide on an input schema for the GI if you are ok with that?

confunguido (approved):

PR #14, opened 2024-11-25
Skeleton of transmission model

ChiragKumar9:
## Description 

This PR implements a skeleton of the transmission model described in `docs/transmission.md`. The focus is not on getting the math right but rather the workflow of how we want to write our transmission model.

## Changes
This implements a simple way to get contacts (with test) and a simple person-to-person transmission based on some generation interval.

## Checklist
- What do you think of the general structuring of functions?
- Places where more test coverage is necessary?
- The overall logic/flow?

### DIFF
--- a/src/contact.rs
+++ b/src/contact.rs
@@ -1,0 +1,105 @@
+use ixa::{
+    context::Context,
+    define_rng,
+    error::IxaError,
+    people::{ContextPeopleExt, PersonId},
+    random::ContextRandomExt,
+};
+use statrs::distribution::Categorical;
+
+use crate::population_loader::Alive;
+
+define_rng!(ContactRng);
+
+pub trait QueryContacts {
+    /// returns an arbitrary contact for the transmitee
+    /// a modeler in the future may institute a more complex contact model
+    /// setting weights by contact setting or shared household or the like in this method
+    fn get_contact(&mut self, transmitee: PersonId) -> Result<Option<PersonId>, IxaError>;
+    /// samples a person from a list of people with a weight by person
+    /// does not check any characteristics about the people
+    /// this is a generic method that is really just used for help sampling the person id
+    /// it is exposed to the user to allow them to make arbitrary sampling decisions
+    fn sample_person_from_list(&mut self, list: Vec<PersonId>, weights: &[f64])
+        -> Option<PersonId>;
+}
+
+impl QueryContacts for Context {
+    fn get_contact(&mut self, transmitee_id: PersonId) -> Result<Option<PersonId>, IxaError> {
+        if self.get_current_population() == 1 {
+            return Err(IxaError::IxaError(
+                "Cannot get a contact when there is only one person in the population.".to_string(),
+            ));
+        };
+        // get list of alive people
+        let alive_people = self.query_people((Alive, true));
+        if alive_people.len() > 1 {
+            // get a random person from the list of alive people
+            // assign weights to each person in the list
+            let mut weights = vec![1.0; alive_people.len()];
+            // get index of transmitee
+            let transmitee_index = alive_people
+                .iter()
+                .position(|&x| x == transmitee_id)
+                .unwrap();
+            // set this weight equal to 0 so we don't select the transmitee as the contact
+            weights[transmitee_index] = 0.0;
+            let contact_id = self.sample_person_from_list(alive_people, &weights);
+            Ok(contact_id)
+        } else {
+            // there are no contacts in the population -- the one remaining alive person must be the transmitee
+            // or there are no alive people in the population
+            Ok(None)
+        }
+    }
+
+    #[allow(clippy::cast_possible_truncation)]
+    #[allow(clippy::cast_sign_loss)]
+    fn sample_person_from_list(
+        &mut self,
+        list: Vec<PersonId>,
+        weights: &[f64],
+    ) -> Option<PersonId> {
+        // subtract one because the weights are 1-indexed
+        let index = self.sample_distr(ContactRng, Categorical::new(weights).unwrap());
+        Some(list[index as usize])
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::QueryContacts;
+    use crate::population_loader::Alive;
+    use ixa::{context::Context, people::ContextPeopleExt, random::ContextRandomExt};
+
+    #[test]
+    #[should_panic(
+        expected = "Cannot get a contact when there is only one person in the population."
+    )]
+    fn test_cant_get_contact_in_pop_of_one() {
+        let mut context = Context::new();
+        let transmitee = context.add_person(()).unwrap();
+        let _ = context.get_contact(transmitee).unwrap();
+    }
+
+    #[test]
+    fn test_return_none() {
+        let mut context = Context::new();
+        context.init_random(108);
+        let transmitee = context.add_person(()).unwrap();
+        let _ = context.add_person((Alive, false)).unwrap();
+        let observed_contact = context.get_contact(transmitee).unwrap();
+        assert!(observed_contact.is_none());
+    }
+
+    #[test]
+    fn test_return_remaining_alive_person() {
+        let mut context = Context::new();
+        context.init_random(108);
+        let transmitee = context.add_person(()).unwrap();
+        let _ = context.add_person((Alive, false)).unwrap();
+        let presumed_contact = context.add_person(()).unwrap();
+        let observed_contact = context.get_contact(transmitee).unwrap();
+        assert_eq!(observed_contact.unwrap(), presumed_contact);
+    }
+}
--- a/src/contact.rs
+++ b/src/contact.rs
@@ -26,8 +26,10 @@ pub trait ContextContactExt {
 #[allow(clippy::cast_possible_truncation)]
 #[allow(clippy::cast_sign_loss)]
 fn sample_person_from_list(context: &mut Context, list: &[PersonId], weights: &[f64]) -> PersonId {
-    let index = context.sample_distr(ContactRng, Categorical::new(weights).unwrap());
-    list[index as usize]
+    let index = context
+        .sample_distr(ContactRng, Categorical::new(weights).unwrap())
+        .floor() as usize;
+    list[index]
 }
 
 impl ContextContactExt for Context {
--- a/src/contact.rs
+++ b/src/contact.rs
@@ -5,7 +5,6 @@ use ixa::{
     people::{ContextPeopleExt, PersonId},
     random::ContextRandomExt,
 };
-use statrs::distribution::Categorical;
 
 use crate::population_loader::Alive;
 
@@ -23,15 +22,6 @@ pub trait ContextContactExt {
     fn get_contact(&mut self, transmitter_id: PersonId) -> Result<Option<PersonId>, IxaError>;
 }
 
-#[allow(clippy::cast_possible_truncation)]
-#[allow(clippy::cast_sign_loss)]
-fn sample_person_from_list(context: &mut Context, list: &[PersonId], weights: &[f64]) -> PersonId {
-    let index = context
-        .sample_distr(ContactRng, Categorical::new(weights).unwrap())
-        .floor() as usize;
-    list[index]
-}
-
 impl ContextContactExt for Context {
     fn get_contact(&mut self, transmitter_id: PersonId) -> Result<Option<PersonId>, IxaError> {
         if self.get_current_population() == 1 {
@@ -44,19 +34,18 @@ impl ContextContactExt for Context {
         // We will sample a random person from this list.
         let mut alive_people = self.query_people((Alive, true));
         if alive_people.len() > 1 {
-            // In the future, we might like to sample people from the list by weights according
-            // to some contact matrix.
-            let mut weights = vec![1.0; alive_people.len()];
+            // Remove the transmitter from the list of contacts.
             // Get the index of the transmitter.
             let transmitter_index = alive_people
                 .iter()
                 .position(|&x| x == transmitter_id)
                 .unwrap();
-            // Remove the transmitter from the list of contacts.
             alive_people.remove(transmitter_index);
-            weights.remove(transmitter_index);
-            let contact_id = sample_person_from_list(self, &alive_people, &weights);
-            Ok(Some(contact_id))
+
+            // In the future, we might like to sample people from the list by weights according
+            // to some contact matrix. We would use sample_weighted instead.
+            let index = self.sample_range(ContactRng, 0..alive_people.len());
+            Ok(Some(alive_people[index]))
         } else {
             // This means that there are no eligible contacts in the population besides the transmitter.
             Ok(None)
--- a/src/contact.rs
+++ b/src/contact.rs
@@ -40,7 +40,7 @@ impl ContextContactExt for Context {
                 .iter()
                 .position(|&x| x == transmitter_id)
                 .unwrap();
-            alive_people.remove(transmitter_index);
+            alive_people.swap_remove(transmitter_index);
 
             // In the future, we might like to sample people from the list by weights according
             // to some contact matrix. We would use sample_weighted instead.
--- a/src/contact.rs
+++ b/src/contact.rs
@@ -11,64 +11,60 @@ use crate::population_loader::Alive;
 
 define_rng!(ContactRng);
 
-pub trait QueryContacts {
-    /// returns an arbitrary contact for the transmitee
-    /// a modeler in the future may institute a more complex contact model
-    /// setting weights by contact setting or shared household or the like in this method
-    fn get_contact(&mut self, transmitee: PersonId) -> Result<Option<PersonId>, IxaError>;
-    /// samples a person from a list of people with a weight by person
-    /// does not check any characteristics about the people
-    /// this is a generic method that is really just used for help sampling the person id
-    /// it is exposed to the user to allow them to make arbitrary sampling decisions
-    fn sample_person_from_list(&mut self, list: Vec<PersonId>, weights: &[f64])
-        -> Option<PersonId>;
+pub trait ContextContactExt {
+    /// Returns a potential contact for the transmitter.
+    /// Returns Ok(None) if there are no eligible contacts.
+    /// In the future, this function can be expanded to return a
+    /// contact specific to the person's household or to weight
+    /// drawing contacts be a contact matrix.
+    ///
+    /// Errors
+    /// - If there is only one person in the population.
+    fn get_contact(&mut self, transmitter_id: PersonId) -> Result<Option<PersonId>, IxaError>;
 }
 
-impl QueryContacts for Context {
-    fn get_contact(&mut self, transmitee_id: PersonId) -> Result<Option<PersonId>, IxaError> {
+#[allow(clippy::cast_possible_truncation)]
+#[allow(clippy::cast_sign_loss)]
+fn sample_person_from_list(context: &mut Context, list: &[PersonId], weights: &[f64]) -> PersonId {
+    let index = context.sample_distr(ContactRng, Categorical::new(weights).unwrap());
+    list[index as usize]
+}
+
+impl ContextContactExt for Context {
+    fn get_contact(&mut self, transmitter_id: PersonId) -> Result<Option<PersonId>, IxaError> {
         if self.get_current_population() == 1 {
             return Err(IxaError::IxaError(
                 "Cannot get a contact when there is only one person in the population.".to_string(),
             ));
         };
-        // get list of alive people
-        let alive_people = self.query_people((Alive, true));
+        // Get list of alive people (for now). May be expanded in the future to instead be
+        // list of alive people in the transmitter's contact setting or household.
+        // We will sample a random person from this list.
+        let mut alive_people = self.query_people((Alive, true));
         if alive_people.len() > 1 {
-            // get a random person from the list of alive people
-            // assign weights to each person in the list
+            // In the future, we might like to sample people from the list by weights according
+            // to some contact matrix.
             let mut weights = vec![1.0; alive_people.len()];
-            // get index of transmitee
-            let transmitee_index = alive_people
+            // Get the index of the transmitter.
+            let transmitter_index = alive_people
                 .iter()
-                .position(|&x| x == transmitee_id)
+                .position(|&x| x == transmitter_id)
                 .unwrap();
-            // set this weight equal to 0 so we don't select the transmitee as the contact
-            weights[transmitee_index] = 0.0;
-            let contact_id = self.sample_person_from_list(alive_people, &weights);
-            Ok(contact_id)
+            // Remove the transmitter from the list of contacts.
+            alive_people.remove(transmitter_index);
+            weights.remove(transmitter_index);
+            let contact_id = sample_person_from_list(self, &alive_people, &weights);
+            Ok(Some(contact_id))
         } else {
-            // there are no contacts in the population -- the one remaining alive person must be the transmitee
-            // or there are no alive people in the population
+            // This means that there are no eligible contacts in the population besides the transmitter.
             Ok(None)
         }
-    }
-
-    #[allow(clippy::cast_possible_truncation)]
-    #[allow(clippy::cast_sign_loss)]
-    fn sample_person_from_list(
-        &mut self,
-        list: Vec<PersonId>,
-        weights: &[f64],
-    ) -> Option<PersonId> {
-        // subtract one because the weights are 1-indexed
-        let index = self.sample_distr(ContactRng, Categorical::new(weights).unwrap());
-        Some(list[index as usize])
     }
 }
 
 #[cfg(test)]
 mod test {
-    use super::QueryContacts;
+    use super::ContextContactExt;
     use crate::population_loader::Alive;
     use ixa::{context::Context, people::ContextPeopleExt, random::ContextRandomExt};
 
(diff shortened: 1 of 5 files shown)


src/transmission_manager.rs
```diff
@@ -0,0 +1,293 @@
...
+define_person_property_with_default!(
+    InfectiousStatusType,
+    InfectiousStatus,
+    InfectiousStatus::Susceptible
+);
+
+fn evaluate_transmission(context: &mut Context, contact_id: PersonId, _transmittee_id: PersonId) {
+    // for now, we assume all transmission events are sucessful
+    // we have the contact_id as an argument to this fcn so that person-person pair transmission potential
+    // based on interventions can be evaluated to determine whether this transmission event is actually successful
```
ekr-cfa, line 32: Please fix spelling and grammar.

src/transmission_manager.rs
```diff
@@ -0,0 +1,293 @@
...
+    if matches!(
```
ekr-cfa, line 33: If you derive PartialEq then you can use ==

src/transmission_manager.rs
```diff
@@ -0,0 +1,293 @@
...
+
+fn infection_attempt(
+    context: &mut Context,
+    transmitee_id: PersonId,
+    gi: f64,
+    num_infection_attempts_remaining: f64,
+    next_infection_time_unif: f64,
+) -> Result<(), IxaError> {
+    // if the person still has people left to infect
+    if num_infection_attempts_remaining >= 1.0 {
```
ekr-cfa, line 53: Why isn't this an integer?

src/transmission_manager.rs
```diff
@@ -0,0 +1,293 @@
...
+        // this is a method from a trait extention implemented in the contact manager
+        // as long as this method returns a contact id, it can use any underlying sampling strategy
```
ekr-cfa, line 55:
```suggestion
        // This is a method from a trait extension implemented in the contact manager.
        // As long as this method returns a contact id, it can use any underlying sampling strategy
```

ekr-cfa (commented):
Please fix all the grammar and spelling in the comments. I flagged a few locations, but this is a global issue.

src/transmission_manager.rs
```diff
@@ -0,0 +1,293 @@
...
+    // we don't care about the other cases here
+    if matches!(event.previous, InfectiousStatus::Susceptible) {
+        // get the number of infection attempts this person will have
+        // ok to use unwrap here because we pass inputs (r_0) through a validator
+        let num_infection_attempts =
+            context.sample_distr(TransmissionRng, Poisson::new(r_0).unwrap());
+        // start scheduling infection attempt events for this person
+        // people who have num_infection_attempts = 0 are still passed through this logic so that
+        // they are set to recovered
+        schedule_next_infection_attempt(context, event.person_id, gi, num_infection_attempts, 0.0);
```
ekr-cfa, line 151: Why are you carrying `gi` around like this rather than either (1) storing it witht he person or (2) getting it from parameters each time?

src/transmission_manager.rs
```diff
@@ -0,0 +1,293 @@
...
+    }
+}
+
+pub fn init(context: &mut Context) {
+    let parameters = context
+        .get_global_property_value(Parameters)
+        .unwrap()
+        .clone();
```
ekr-cfa, line 159: I don't think you need to clone here. Instead, make temporary variables for r_0 and generation_interval, which are Copy, and capture those.

src/transmission_manager.rs
```diff
@@ -0,0 +1,293 @@
...
+}
+
+#[cfg(test)]
+mod test {
+    use crate::{parameters::ParametersValues, population_loader::Alive};
+
+    use super::*;
+    use ixa::{context::Context, random::ContextRandomExt};
+
+    fn set_up(r_0: f64) -> Context {
```
ekr-cfa, line 179: I would call this `setup`

src/transmission_manager.rs
```diff
@@ -0,0 +1,293 @@
...
+        let mut context = Context::new();
+        context.init_random(p_values.seed);
+        context.set_global_property_value(Parameters, p_values);
+        context
+    }
+
+    #[test]
+    fn test_transitions() {
+        // set a super small r_0 so that the person has a very low probability of infecting others
+        let mut context = set_up(0.000_000_000_000_000_01);
```
ekr-cfa, line 198: Why does this matter? You only have one person in the simulation, so even if you were definitely going to infect someone, the simulation should terminate the same way

ekr-cfa (changes requested):
I need to finish reviewing the tests, which I will do tomorrow.

src/transmission_manager.rs
ChiragKumar9: Thanks. Currently, the skeleton of the `get_contact` method in this PR raises an error when trying to query a contact in a population of one person. So, if we set $r_0 = 1$, the agent would query a contact, and an error would be raised.

I implemented the `get_contact` this way because to me trying to query a contact in a population of one means something is wrong with your simulation set up and should raise an error. On the other hand, being unable to find an eligible contact should return `None`, and I wanted a way to differentiate between the two. However, for now we can certainly group these two together while we are just trying to get a skeleton in place.

src/transmission_manager.rs
ChiragKumar9: Annoyingly, randomly generating from `Pois` returns a float instead of an integer even though all sampled values are integral. Any thoughts on how to fix this without casting/is casting just the right solution here?

src/contact.rs
```diff
@@ -0,0 +1,105 @@
...
+            // get a random person from the list of alive people
+            // assign weights to each person in the list
+            let mut weights = vec![1.0; alive_people.len()];
+            // get index of transmitee
+            let transmitee_index = alive_people
+                .iter()
+                .position(|&x| x == transmitee_id)
+                .unwrap();
+            // set this weight equal to 0 so we don't select the transmitee as the contact
+            weights[transmitee_index] = 0.0;
```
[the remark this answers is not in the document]
ChiragKumar9: Thanks. I removed the transmitter (you are right there too with the nomenclature -- we mixed up transmittee and transmitter). However, I do think we should keep the weights. In the future, we will want to sample from the list of eligible contacts potentially with weights based on person characteristics.

src/transmission_manager.rs
jasonasher: I think casting is the solution here: `rng.sample(Poisson) as usize` for example

src/contact.rs
```diff
@@ -11,64 +11,60 @@ use crate::population_loader::Alive;
...
+    fn get_contact(&mut self, transmitter_id: PersonId) -> Result<Option<PersonId>, IxaError>;
 }
 
-impl QueryContacts for Context {
-    fn get_contact(&mut self, transmitee_id: PersonId) -> Result<Option<PersonId>, IxaError> {
+#[allow(clippy::cast_possible_truncation)]
+#[allow(clippy::cast_sign_loss)]
+fn sample_person_from_list(context: &mut Context, list: &[PersonId], weights: &[f64]) -> PersonId {
+    let index = context.sample_distr(ContactRng, Categorical::new(weights).unwrap());
+    list[index as usize]
```
ekr-cfa, line 30:
```suggestion
    let index = context.sample_distr(ContactRng, Categorical::new(weights).unwrap()) as usize;
    list[index]
```

Technically speaking, the index is an integer not an f64.

ekr-cfa: Have you checked in the doc that this is guaranteed to be floor()? Perhaps floor() first would be safer.

src/contact.rs
```diff
@@ -11,64 +11,60 @@ use crate::population_loader::Alive;
...
+                .position(|&x| x == transmitter_id)
                 .unwrap();
-            // set this weight equal to 0 so we don't select the transmitee as the contact
-            weights[transmitee_index] = 0.0;
-            let contact_id = self.sample_person_from_list(alive_people, &weights);
-            Ok(contact_id)
+            // Remove the transmitter from the list of contacts.
+            alive_people.remove(transmitter_index);
+            weights.remove(transmitter_index);
+            let contact_id = sample_person_from_list(self, &alive_people, &weights);
```
ekr-cfa, line 56: Now you don't need weights, so why don't you just sample from the uniform distribution of (0..alive_people) and then floor, inline, rather than having a separate function.

src/transmission_manager.rs
```diff
@@ -26,14 +31,15 @@ define_person_property_with_default!(
...
-    if matches!(
-        context.get_person_property(contact_id, InfectiousStatusType),
-        InfectiousStatus::Susceptible
-    ) {
+fn evaluate_transmission(context: &mut Context, contact_id: PersonId, _transmitter_id: PersonId) {
+    // For now, we assume all transmission events are sucessful.
+    // However, in the future, the success of a transmission event may depend on person-level interventions,
+    // such as whether an agent is wearing a mask. For this reason, we pass the transmitter as well to this
+    // function. This will allow us to evaluate the success of a transmission event based on the properties of
+    // both the transmitter and the contact in the future.
```
ekr-cfa, line 39: Shouldn't this comment be above the function? And explain what the function does first.

src/transmission_manager.rs
```diff
@@ -44,36 +50,38 @@ fn evaluate_transmission(context: &mut Context, contact_id: PersonId, _transmitt
...
-    // if the person still has people left to infect
-    if num_infection_attempts_remaining >= 1.0 {
-        // this is a method from a trait extention implemented in the contact manager
-        // as long as this method returns a contact id, it can use any underlying sampling strategy
-        let contact_id = context.get_contact(transmitee_id)?;
-        // evaluate transmission is its own function because there will be
-        // intervention-based logic there eventually
-        // if there are no contacts, to infect, do nothing
+    // Only attempt the infection if the person has infection attempts remaining.
+    if num_infection_attempts_remaining >= 1 {
```
ekr-cfa, line 58:
```suggestion
    if num_infection_attempts_remaining > 0 {
```
This is now an integer.

src/transmission_manager.rs
```diff
@@ -44,36 +50,38 @@ fn evaluate_transmission(context: &mut Context, contact_id: PersonId, _transmitt
...
+        // This is a method from a trait extension implemented in `mod contact`.
+        // As long as the method returns a contact id, it can use any underlying sampling strategy
+        // to obtain that contact, and that strategy can be separately implemented in `mod contact`
+        // without changing the logic here.
+        let contact_id = context.get_contact(transmitter_id)?;
+
+        // We evvaluate transmission in its own function because there will be eventually
```
ekr-cfa, line 65:
```suggestion
        // We evaluate transmission in its own function because there will be eventually
```

src/transmission_manager.rs
```diff
@@ -44,36 +50,38 @@ fn evaluate_transmission(context: &mut Context, contact_id: PersonId, _transmitt
...
+        // be intervention-based logic there.
+        // If there are no contacts to infect, do nothing.
         if let Some(contact_id) = contact_id {
-            evaluate_transmission(context, contact_id, transmitee_id);
+            evaluate_transmission(context, contact_id, transmitter_id);
```
ekr-cfa, line 69: Put the above comment about why there is a separate function here and leave the conditional comment above.

src/transmission_manager.rs
```diff
@@ -44,36 +50,38 @@ fn evaluate_transmission(context: &mut Context, contact_id: PersonId, _transmitt
...
         }
-        // schedule the subsequent infection attempt for this infected agent,
-        // which happens at a greater value of the GI
+
+        // Schedule the next infection attempt for this infected agent.
+        // The next infection attempt happens at a greater value of the GI than the last infection attempt.
```
ekr-cfa, line 73: I don't know what this sentence is saying.

src/transmission_manager.rs
```diff
@@ -82,48 +90,45 @@ fn infection_attempt(
...
     context: &mut Context,
-    transmitee_id: PersonId,
-    gi: f64,
-    num_infection_attempts_remaining: f64,
+    transmitter_id: PersonId,
+    num_infection_attempts_remaining: usize,
     last_infection_time_unif: f64,
 ) {
-    // get next infection attempt time
+    // Get the next infection attempt time, which is greater than the last infection attempt time.
```
ekr-cfa, line 120: I would just say "Get the next infection time". The greater than part is enforced elsewhere.

src/transmission_manager.rs
```diff
@@ -132,51 +137,49 @@ fn schedule_next_infection_attempt(
...
 ) {
-    // is the person going from S --> I?
-    // we don't care about the other cases here
-    if matches!(event.previous, InfectiousStatus::Susceptible) {
-        // get the number of infection attempts this person will have
-        // ok to use unwrap here because we pass inputs (r_0) through a validator
+    // Is the person going from S --> I?
+    // We don't care about the other transitions here, but this function will still be triggered
+    // because it watches for any change in InfectiousStatusType.
+    if event.previous == InfectiousStatus::Susceptible {
```
ekr-cfa, line 149: I would assert that the current state is infected.

src/transmission_manager.rs
```diff
@@ -44,36 +50,38 @@ fn evaluate_transmission(context: &mut Context, contact_id: PersonId, _transmitt
...
-            transmitee_id,
-            gi,
-            num_infection_attempts_remaining - 1.0,
+            transmitter_id,
+            num_infection_attempts_remaining - 1,
             next_infection_time_unif,
         );
     }
-    // if no more infection attempts remaining, set the person to recovered
+    // If there are no more infection attempts remaining, set the person to recovered.
```
ekr-cfa, line 81: I would put this case first, and then you can return right away, rather than having an else.

src/transmission_manager.rs
```diff
@@ -214,65 +219,64 @@ mod test {
...
+        // Using some math, we can test that the observed time of the end of the simulation
+        // is what we expect it to be given the number of infection attempts.
+
+        // In our current naive implementation, the expected end time of the simulation is
+        // the sum of the exponential inverse_cdf evaluated at each draw of the uniform distribution.
+
         let mut sum_end_times = 0.0;
-        // r_0 not in this test is meaningless
-        let context = set_up(1.0);
+        // r_0 not in this test is meaningless because we manually set the number of infection attempts.
```
ekr-cfa, line 262: I don't understand what this sentence says.

src/transmission_manager.rs
```diff
@@ -281,13 +285,14 @@ mod test {
             );
         }
         // TODO: figure out way to check math for n_attempts > 1
```
ekr-cfa, line 287: Are you going to address this TODO.

Generally, I would prefer we use `TODO(<id>)` rather than just `TODO`. E.g., `TODO(cym4@cdc.gov)` or `TODO(cym4)`

src/transmission_manager.rs
```diff
@@ -281,13 +285,14 @@ mod test {
...
-        let expected_time_elapsed = f64::powf(params.generation_interval, n_attempts);
+        let expected_time_elapsed =
+            f64::powi(params.generation_interval, n_attempts.try_into().unwrap());
 
         assert!(((sum_end_times / f64::from(n_iter)) - expected_time_elapsed).abs() < 0.1);
     }
 
     #[test]
     fn test_variable_number_of_infection_attempts() {
```
ekr-cfa, line 295: Is there a reason not to just have the function inlined?

ekr-cfa (changes requested):
Thanks for providing some additional comments. Let me see if I can understand the logic here.

Once a person is infected, you draw a random number of people they will infect I (which may be 0) and then:

1. Generate a new infection time T
2. If I < 0, mark them as recovered
3. If I > 0
    * Try to infect someone
    * Decrement I by 1
    * Go to 1
    
Is this correct? And based on the comment in line 155-159, you're doing it this way to mark people recovered in only one place.

This seems like it has the somewhat odd side effect that you recover T seconds after the last time you have infected someone. Is that what's desired?

In any case, this is kind of an unusual structure, so I would add an overall comment describing the structure.

src/transmission_manager.rs
```diff
@@ -0,0 +1,295 @@
...
+        // Get the number of infection attempts this person will have.
+        let r_0 = context.get_global_property_value(Parameters).unwrap().r_0;
+        let num_infection_attempts =
+            context.sample_distr(TransmissionRng, Poisson::new(r_0).unwrap());
+
+        // Start scheduling infection attempt events for this person.
+        // People who have num_infection_attempts = 0 are still passed through this
+        // logic but don't infect anyone. This is so that so that there is only one
+        // place where we need to handle setting people to recovered.
+        schedule_next_infection_attempt(context, event.person_id, num_infection_attempts, 0.0);
```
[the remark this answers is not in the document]
ChiragKumar9: Yes

src/transmission_manager.rs
```diff
@@ -0,0 +1,295 @@
...
+    Exp::new(1.0 / gi)
+        .unwrap()
+        .inverse_cdf(next_infection_time_unif)
+}
+
+fn schedule_next_infection_attempt(
+    context: &mut Context,
+    transmitter_id: PersonId,
+    num_infection_attempts_remaining: f64,
+    last_infection_time_unif: f64,
```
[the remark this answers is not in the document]
ChiragKumar9: Thanks. Cleaned up the notation.

src/transmission_manager.rs
```diff
@@ -0,0 +1,295 @@
...
+) {
+    // Get the next infection attempt time, which is greater than the last infection attempt time.
+    let next_infection_time_unif = get_next_infection_time_unif(context, last_infection_time_unif);
+    let next_infection_time_gi = get_next_infection_time_from_gi(context, next_infection_time_unif);
```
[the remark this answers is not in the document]
ChiragKumar9: Thanks. Good point. Cleaned up function.

src/transmission_manager.rs
```diff
@@ -0,0 +1,295 @@
...
+}
+
+fn handle_infectious_status_change(
+    context: &mut Context,
+    event: PersonPropertyChangeEvent<InfectiousStatusType>,
+) {
+    // Is the person going from S --> I?
+    // We don't care about the other transitions here, but this function will still be triggered
+    // because it watches for any change in InfectiousStatusType.
+    if event.previous == InfectiousStatus::Susceptible {
```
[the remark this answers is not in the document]
ChiragKumar9: Thanks. This is a good catch. Fixed.

src/transmission_manager.rs
ChiragKumar9: Thanks. I use the function multiple times to test multiple numbers of infection attempts now that I added in the proper math to get the expected time.

ChiragKumar9:
> Thanks for providing some additional comments. Let me see if I can understand the logic here.
> 
> Once a person is infected, you draw a random number of people they will infect I (which may be 0) and then:
> 
> 1. Generate a new infection time T
> 2. If I < 0, mark them as recovered
> 3. If I > 0
>    
>    * Try to infect someone
>    * Decrement I by 1
>    * Go to 1
> 
> Is this correct? And based on the comment in line 155-159, you're doing it this way to mark people recovered in only one place.
> 
> This seems like it has the somewhat odd side effect that you recover T seconds after the last time you have infected someone. Is that what's desired?
> 
> In any case, this is kind of an unusual structure, so I would add an overall comment describing the structure.

Thanks, @ekr-cfa. You are absolutely right. After some discussion, I've revised the flow to get rid of the odd side effect you point out. Now, if someone has zero secondary infection attempts, they go straight to recovery. I still keep the code structure I want of only needing to check once.

src/transmission_manager.rs
```diff
@@ -0,0 +1,295 @@
...
+#[derive(Debug, Hash, Eq, PartialEq, Clone, Copy)]
+pub enum InfectiousStatus {
+    Susceptible,
+    Infectious,
+    Recovered,
+}
+
+define_rng!(TransmissionRng);
+
+define_person_property_with_default!(
```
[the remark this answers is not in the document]
ChiragKumar9: In the examples, we generally use `Type` to denote the property name and `InfectiousStatus` to denote the enum. But, I changed it up here to see what the pattern was like. I don't feel strongly. Would be curious what others think.

src/transmission_manager.rs
```diff
@@ -0,0 +1,341 @@
...
+            move |context| {
+                infection_attempt(context, transmitter_id)
+                    .expect("Error finding contact in infection attempt");
+                // Schedule the next infection attempt for this infected agent
+                // once the last infection attempt is over.
+                schedule_next_infection_attempt(
+                    context,
+                    transmitter_id,
+                    num_infection_attempts_remaining - 1,
+                    next_infection_times.uniform,
```
ekr-cfa, line 127: I think if  you change this to next_infection_times_uniform and make a temporary, you can remove the `move` because these scalar values will have Copy.

src/transmission_manager.rs
```diff
@@ -0,0 +1,341 @@
...
+        );
+    }
+}
+
+/// Calculate the next infection time. Draws an increasing value of a uniform distribution
+/// and passes it through the inverse CDF of the generation interval to get the next infection time.
+fn get_next_infection_time(
+    context: &mut Context,
+    last_infection_time_uniform: f64,
+) -> InfectionTimes {
```
ekr-cfa, line 139: I wonder if it would be better to return a tuple here rather than have a type.

src/transmission_manager.rs
```diff
@@ -0,0 +1,341 @@
...
+}
+
+/// This function is called when an infected agent has an infection event scheduled.
+/// This function includes identifying a potential contact to infect, evaluating whether
+/// the transmission event is successful, and scheduling the next infection event.
+fn infection_attempt(context: &mut Context, transmitter_id: PersonId) -> Result<(), IxaError> {
+    // This is a method from a trait extension implemented in `mod contact`.
+    // As long as the method returns a contact id, it can use any underlying sampling strategy
+    // to obtain that contact, and that strategy can be separately implemented in `mod contact`
+    // without changing the logic here.
```
ekr-cfa, line 168: I'm not sure I understand this comment. I tend to think that the note about there being no guarantee about the strategy should go with the definition.

Per our previous discussion, do we want to have this be a trait extension or a function.

src/transmission_manager.rs
```diff
@@ -0,0 +1,341 @@
...
+    let contact_id = context.get_contact(transmitter_id)?;
+
+    // If there are no contacts to infect, do nothing.
+    if let Some(contact_id) = contact_id {
```
ekr-cfa, line 172: Given that you don't use `contact_id` anywhere else, do you want to merge line 169 here,

src/transmission_manager.rs
```diff
@@ -0,0 +1,341 @@
...
+
+    Ok(())
+}
+
+/// Evaluates whether a transmission event is successful based on the characteristics
+/// of the transmitter and the contact. For now, we assume all transmission events are sucessful.
+fn evaluate_transmission(context: &mut Context, contact_id: PersonId, _transmitter_id: PersonId) {
+    // In the future, the success of a transmission event may depend on person-level interventions,
+    // such as whether an agent is wearing a mask. For this reason, we pass the transmitter as well to this
+    // function. This will allow us to evaluate the success of a transmission event based on the properties of
```
ekr-cfa, line 186: allow us in the future, right?

src/transmission_manager.rs
```diff
@@ -0,0 +1,341 @@
...
+    };
+
+    use super::{init, InfectiousStatus, InfectiousStatusType};
+    use ixa::{
+        context::Context, global_properties::ContextGlobalPropertiesExt, people::ContextPeopleExt,
+        random::ContextRandomExt,
+    };
+
+    fn setup(r_0: f64) -> Context {
+        let p_values = ParametersValues {
```
ekr-cfa, line 216: Nit: I would say `params`.

src/transmission_manager.rs
```diff
@@ -0,0 +1,341 @@
...
+
+        // Expected time elapsed comes from the memorylessness of the exponential distribution.
+        let expected_time_elapsed =
+            params.generation_interval * (n_attempts as f64) * ((n_attempts as f64) + 1.0) / 2.0;
+
+        println!(
+            "Expected time elapsed: {}, Observed time elapsed: {}",
+            expected_time_elapsed,
+            sum_end_times / f64::from(n_iter)
+        );
```
ekr-cfa, line 330: Normally you don't `println!()` in tests.

ekr-cfa (changes requested):

src/transmission_manager.rs
ChiragKumar9: > Per our previous discussion, do we want to have this be a trait extension or a function.

In this case, I'm totally agnostic. It would be a private trait, right? In that case, I don't see any difference to what's implemented right now, and I am happy with whatever you or @k88hudson-cfa think is the best style-wise.

src/transmission_manager.rs
ChiragKumar9: Sorry, I couldn't get this to work. I may not be understanding your suggestion correctly. Did you mean the following?

```rust
let next_infection_time_uniform = next_infection_times.0;
context.add_plan(context.get_current_time() + next_infection_times.1, |context| {
    schedule_next_infection_attempt(..., next_infection_time_uniform);
}

```

src/transmission_manager.rs
```diff
@@ -0,0 +1,341 @@
...
+    context.add_plan(0.0, |context| {
+        seed_infections(context);
+    });
+}
+
+/// This function seeds the initial infections in the population.
+fn seed_infections(context: &mut Context) {
+    // For now, we just pick a random person and make them infectious.
+    // In the future, we may pick people based on specific person properties.
+    let alive_people = context.query_people((Alive, true));
```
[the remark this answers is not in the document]
ChiragKumar9: Great. I plan on switching once that PR lands.

src/transmission_manager.rs
```diff
@@ -106,37 +101,37 @@ fn schedule_next_infection_attempt(
...
-                    num_infection_attempts_remaining - 1,
-                    next_infection_times.uniform,
-                );
-            },
-        );
+        return;
     }
+
+    // Get the next infection attempt time.
+    let next_infection_times = get_next_infection_time(context, last_infection_time_uniform);
```
ekr-cfa, line 108: If you instead do:

`let (next_infection_time_unif, next_infection_time_gi) = ....`

Then:

1. It will be clearer
2. You will not need to do .0 and .1
3. You will not need `move` on line 117

ekr-cfa (approved):
I have one small suggestion for improvement.

src/transmission_manager.rs
ChiragKumar9: Thanks! I didn't see this. Agree on improving clarity, but I can't get compilation without the `move` in the closure.

PR #15, opened 2024-11-25
Gce minimum report

confunguido:
Reopening from previous closed PR #10, which provides a skeleton for a report.

### DIFF
--- a/src/periodic_report_population.rs
+++ b/src/periodic_report_population.rs
@@ -81,7 +81,7 @@ pub fn init(context: &mut Context, output_dir: &Path) -> Result<(), IxaError> {
         update_property_set(context, event);
     });
 
-    context.add_report::<PersonReportItem>(&parameters.output_file)?;
+    context.add_report::<PersonReportItem>(&parameters.population_periodic_report)?;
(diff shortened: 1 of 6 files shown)


src/periodic_report_population.rs
```diff
@@ -0,0 +1,89 @@
...
+    }
+);
+
+create_report_trait!(PopulationReportItem);
+
+fn send_population_report(context: &mut Context) {
+    let population_data = context.get_data_container(PopulationReportPlugin).unwrap();
+
+    let current_census_set = &population_data.census_tract_set;
+    for age_it in 0..100 {
```
ekr-cfa, line 44: What does `age_it` stand for?

ekr-cfa (approved):
This seems right, but I think @k88hudson-cfa  is pretty close to having a generic periodic reports mechanism, so I wonder if we should wait for that.

src/periodic_report_population.rs
confunguido: It stands for age_iterator but this is a bad name, maybe. I can change that.

PR #18, opened 2024-11-27
Simple synthetic population generator 

confunguido:
This PR includes a script to generate synthetic populations. The algorithm samples houses from the census datasets available in pums based on the household weight and the desired population size. Region coordinates are downloaded for each region based on their PUMA codes. These coordinates are then assigned to each house based on their PUMA code. Finally, two files are generated: people (age, home_id), and region (region_id, lat, lon). Region id is generated as follows: 2 digits for state, 9 digits for PUMA code. Home id is region id + 6 digits that represent the home number.

Census tracts have been changed to region id in the ixa code. 

These populations are for testing purposes only.

### DIFF
--- a/scripts/create_synthetic_population.R
+++ b/scripts/create_synthetic_population.R
@@ -1,0 +1,106 @@
+## =================================#
+## Setup ---------------
+## =================================#
+library(tidyverse)
+library(tigris)
+library(tidycensus)
+library(patchwork)
+
+set.seed(1234)
+
+state_synth <- "WY"
+year_synth <- 2023
+population_size <- 1000
+
+## =================================#
+## Get population ---------------
+## =================================#
+pums_vars <- pums_variables |>
+  filter(year == 2018, survey == "acs1") |>
+  distinct(var_code, var_label, data_type, level)
+
+person_variables <- c(
+  "SPORDER", "SERIALNO", "PWGTP",
+  "AGEP", "SEX", "PUMA", "REGION"
+)
+house_variables <- c("WGTP", "NP")
+
+sample_pums <- get_pums(
+  variables = c(person_variables, house_variables),
+  state = state_synth,
+  survey = "acs1",
+  year = year_synth
+)
+
+household_pums <- sample_pums |>
+  dplyr::select(SERIALNO, all_of(house_variables)) |>
+  distinct()
+
+## =================================#
+## Create population ---------------
+## =================================#
+synth_pop_df <- tibble()
+house_counter <- 0
+while (nrow(synth_pop_df) < population_size) {
+  house_counter <- house_counter + 1
+  synth_pop_df <- bind_rows(synth_pop_df, household_pums) |>
+    sample_n(1, weight = WGTP) |>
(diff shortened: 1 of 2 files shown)


src/population_loader.rs
```diff
@@ -11,25 +11,25 @@ use std::path::PathBuf;
 #[allow(non_snake_case)]
```
KOVALW, line 11: We shouldn't need this clippy exception now

scripts/create_synthetic_population.R
```diff
@@ -0,0 +1,92 @@
...
+## =================================#
+library(tidyverse)
+library(tigris)
+library(tidycensus)
+
+state_synth <- "WY"
+year_synth <- 2023
+population_size <- 100
+
+## =================================#
```
KOVALW, line 12: We should probably have a fixed seed here to have a consistent synthetic population

KOVALW:
Why is the build and test workflow failing here? The branch run works locally for me

scripts/create_synthetic_population.R
```diff
@@ -0,0 +1,95 @@
...
+
+synth_pop_region_df <- synth_pop_df |>
+  left_join(
+    pumas_st |>
+      dplyr::select(STATEFP20, PUMACE20, INTPTLAT20, INTPTLON20),
+    by = c("PUMA" = "PUMACE20")
+  ) |>
+  dplyr::select(-geometry) |>
+  mutate(
+    region_id = sprintf("%02d%09d", as.numeric(STATE), as.numeric(PUMA)),
```
KOVALW, line 66: Even though region ID isn't in the ASPR set, so we wouldn't name it in camel case for consistency across implementations, should it be in camel case to stylistically match `homeId`?

scripts/create_synthetic_population.R
```diff
@@ -0,0 +1,106 @@
...
+## =================================#
+synth_pop_df <- tibble()
+house_counter <- 0
+while (nrow(synth_pop_df) < population_size) {
+  house_counter <- house_counter + 1
+  synth_pop_df <- bind_rows(synth_pop_df, household_pums) |>
+    sample_n(1, weight = WGTP) |>
+    left_join(sample_pums, by = (c("SERIALNO", "WGTP", "NP"))) |>
+    mutate(house_number = house_counter)
+}
```
KOVALW, line 50: Does it matter that this can produce a data frame that doesn't necessarily have exactly the `population_size` but only at least that size? I don't think so since we can calculate the population later when using the synthetic data, but wanted to draw attention to the potential variability. I could really only see this mattering for population sizes that are small enough to be within the distribution of residents per household

confunguido: If we get fancier, maybe we can fix that. But for how simple this is, I don't think we should care too much about that.

scripts/create_synthetic_population.R
confunguido: You're right

ChiragKumar9 (approved):
Good with me as a first pass

PR #20, opened 2024-11-29
Add command line argument to overwrite reports

ChiragKumar9:
No rush on this PR. Can even merge into main after we've merged in Will's pop loader.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -20,6 +20,9 @@ struct Args {
     /// path to the output directory
     #[arg(short, long)]
     output_directory: PathBuf,
+
+    #[arg(short, long, default_value_t = false, default_missing_value = "true")]
+    force_overwrite: bool,
 }
 
 mod periodic_report_population;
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -20,6 +20,9 @@ struct Args {
     /// path to the output directory
     #[arg(short, long)]
     output_directory: PathBuf,
+
+    #[arg(short, long, default_value = "false", default_missing_value = "true")]
```
confunguido, line 24: whats the difference between default_value and default_missing_value?

confunguido (approved):
Looks good to me, but not sure I understand the default values for the overwrite.

src/main.rs
ChiragKumar9: `default_value` is the default value of the `force_overwrite` attribute of the struct when nothing is provided by the user. `default_missing_value` is the default value when the user passes the `-f` or `--force-overwrite` flag to the program but nothing else afterwards, so no `-f true` or `-f false` (which the user could do and is maybe confusing behavior).

PR #21, opened 2024-11-29
Suggesting some changes to writing

confunguido:
Some suggestions for writing. Will check math later. The document looks good but I think it's too long. I suggested changes to be more concise and re-ordered some paragraphs.


ChiragKumar9:
Thanks, @confunguido! I incorporated these comments into my branch. Would appreciate your review there again. If you don't mind, I will close this PR.

PR #23, opened 2024-12-04
Replace call to get_person_id

ekr-cfa:


KOVALW (approved):
LGTM and works fine on my local

PR #25, opened 2024-12-05
A simple report of infectious prevalence

ChiragKumar9:
We will add a report for `HealthStatus` once that is implemented in the model.

It's my understanding that we plan to do analysis in a separate repo, so any scripts to post-process the model outputs will go there. I am including an R script just to make a quick plot of the outputs, but I can remove this if that should go in the other repo.

Thanks!

### DIFF
--- a/scripts/plot_outputs.R
+++ b/scripts/plot_outputs.R
@@ -1,0 +1,45 @@
+## ===============================#
+## Setup ---------
+## ===============================#
+
+library(tidyverse)
+ggplot2::theme_set(ggplot2::theme_classic())
+
+## ===============================#
+## Read person properties reports
+## ===============================#
+
+demographic_report <- readr::read_csv(file.path(
+  "output",
+  "person_demographics_count.csv"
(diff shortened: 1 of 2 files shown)


scripts/plot_outputs.R
```diff
@@ -0,0 +1,45 @@
+## ===============================#
+## Setup ---------
+## ===============================#
+
+library(tidyverse)
```
ekr-cfa, line 5: Nit: is the convention really to have a blank line after the block comments?

ekr-cfa (approved):
LGTM

scripts/plot_outputs.R
ChiragKumar9: According to ChatGPT, leaving a blank line is the right convention. I couldn't find it in the official style guide though -- the closest I came to was this: https://medium.com/@sdshwetadixit/comments-and-headers-in-r-a-guide-for-all-2faae3bdc65c.

PR #27, opened 2024-12-09
Gce report fix

confunguido:
Merging both reports into one. I'm also removing unused parameters related to the reports.

### DIFF
--- a/scripts/plot_outputs.R
+++ b/scripts/plot_outputs.R
@@ -23,6 +23,6 @@ infectious_report |>
   group_by(t, InfectiousStatus) |>
   summarise(count = sum(count), .groups = "drop") |>
   ggplot(aes(x = t, y = count)) +
-  geom_line(aes(color =InfectiousStatus)) +
+  geom_line(aes(color = InfectiousStatus)) +
   xlab("Day") +
   ylab("Number of people")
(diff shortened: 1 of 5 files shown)


scripts/plot_outputs.R
```diff
@@ -9,11 +9,6 @@ ggplot2::theme_set(ggplot2::theme_classic())
...
 ## ===============================#
 
-demographic_report <- readr::read_csv(file.path(
-  "output",
-  "person_demographics_count.csv"
-))
-
 infectious_report <- readr::read_csv(file.path(
   "output",
   "person_infectious_count.csv"
```
ChiragKumar9, line 14: The report here is called the `person_infectious_count` whereas in `main` it's called `person_property_count`. Please update the report name here.

ChiragKumar9 (changes requested):

KOVALW (approved):
LGTM apart from Chirag's catch on the `person_property_count` variable name

scripts/plot_outputs.R
confunguido: I'm very curious of how hat even happened. We plotted something and "person_infectious_count.csv" does not exist on my computer.

ChiragKumar9 (approved):
LGTM, thanks.

PR #29, opened 2024-12-10
Introduce a natural history manager to sample infection attempts from a user-specified GI trajectory

ChiragKumar9:
This PR (now revamped!) introduces:

1. A proposal for a general structure for inputting, storing and querying natural history parameters.
2. Applying that input schema to values of the generation interval over time since an agent is infected.
3. The ability to query the value of the inverse generation interval (with smooth interpolation using a cubic spline) for scheduling infection attempts.

Very open to feedback. The main concern I have is about my proposal for modeling natural history parameters that are correlated -- like the generation interval and symptom improvement time as we have in our COVID isolation guidance work. I propose a way to do this, but I want to ensure that I am not trying to basically "throw all parameters together and hope for the best".


docs/viral-load-infectiousness.md
```diff
@@ -0,0 +1,23 @@
+# Infectiousness over time
+
+## Overview
+
+We assume that infectiousness over time varies with the viral load, similar to how we modeled infectiouness over time in our isolation guidance modeling response.
```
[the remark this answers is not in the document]
confunguido: I would like to have some guidance on this. It is a bit difficult to edit someone else's markdown when lines are broken down like that, but I see the benefit of breaking them up for track changes. I just don't know if you have any suggestions on how to edit this, or maybe some good emacs setup.

src/natural_history_manager.rs
```diff
@@ -0,0 +1,572 @@
...
+            // linear extrapolation using the last two values in `xs`.
+            let traj_len = xs.len();
+            return linear_interpolation(
+                xs[traj_len - 2],
+                xs[traj_len - 1],
+                ys[traj_len - 2],
+                ys[traj_len - 1],
+                xp,
+            );
+        }
```
ekr-cfa, line 233: What happens if someone provides `xp` really far in the future? Won't you extrapolate above 1?

src/natural_history_manager.rs
```diff
@@ -0,0 +1,572 @@
...
+        Some(i) => match i {
```
ekr-cfa, line 234: Can you do flatten this and do `Some(1)` and `_` as the arms?

src/natural_history_manager.rs
```diff
@@ -0,0 +1,572 @@
...
+            // The index can never return 0 because we handle that case at the beginning of the function.
+            _ => i,
+        },
+    };
+    // If our interpolation point is in the range of the `xs` samples, we can use cubic spline interpolation.
+    cubic_spline_interpolation(
+        xs[(upper_window_index - 2)..=(upper_window_index + 1)]
+            .try_into()
+            .unwrap(),
+        // Since `upper_window_index` is the third index in the trajectory, we need to subtract 2 to get the first index.
```
ekr-cfa, line 247: Shouldn't this comment be on line 244?

src/natural_history_manager.rs
```diff
@@ -0,0 +1,572 @@
...
+    )
+}
+
+/// Fits a line between two points and returns the value of the line at a given x-value `xp`.
+/// Sensibly handles equal x-values by returning `y1` for `xp < x1`, `y2` for `xp > x2`, and
+/// `(y1 + y2) / 2` for `xp = x1 = x2`.
+fn linear_interpolation(x1: f64, x2: f64, y1: f64, y2: f64, xp: f64) -> f64 {
+    // At the tails of the CDF, the CDF moves very slowly, so with numerical imprecision, x2 may equal x1.
+    #[allow(clippy::float_cmp)]
+    if x2 == x1 {
```
ekr-cfa, line 261: I'm trying to figure out how you get into this state. By construction, isn't `xp < x2`?

src/natural_history_manager.rs
```diff
@@ -0,0 +1,572 @@
...
+            return y2;
+        }
+        return (y1 + y2) / 2.0;
+    }
+    (y2 - y1) / (x2 - x1) * (xp - x1) + y1
+}
+
+/// Fits a cubic spline between four points and returns the value of the spline at a given x-value `xp`.
+/// Requires that `xp` be between the second and third ordered points in the `xs` array.
+fn cubic_spline_interpolation(xs: &[f64; 4], ys: &[f64; 4], xp: f64) -> f64 {
```
ekr-cfa, line 275: I'm not a math guy but given the expected density of these points, is cubic really that much better than linear? 

My instinct would actually be to do all this interpolation once at loading time down to some precision limit (.1 days?) and then just look up the closest value and take it.

src/natural_history_manager.rs
```diff
@@ -0,0 +1,572 @@
...
+        interpolate(gi_trajectory, times, gi_cdf_value_unif)
+    }
+}
+
+/// An interpolation routine that expects a paired set of values `xs` and `ys` that represent samples
+/// from a given function. The function is evaluated at a given x value `xp` using cubic spline interpolation
+/// when there are at least two samples above and below `xp`. Otherwise, it uses linear extrapolation at the tails.
+/// Assumes that function samples are sorted so that the `xs` are in ascending order.
+fn interpolate(xs: &[f64], ys: &[f64], xp: f64) -> f64 {
+    let upper_window_index_option = xs.iter().position(|&x| x > xp);
```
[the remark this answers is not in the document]
ekr-cfa: Can you do binary search here?

ekr-cfa (changes requested):
I went over the main logic of the code and made some comments. I haven't reviewed the tests yet, but I think this could use another spin.

You should also spellcheck the comments.

docs/natural-history-inputs.md
```diff
@@ -0,0 +1,95 @@
...
+| 0 | 1.0 | 0.5 |
+| 0 | 2.0 | 1.0 |
+| 1 | 0.0 | 0.0 |
+| 1 | 1.0 | 0.3 |
+| ... | ... | ... |
+
+The `id` is an identifier that marks a distinct sample from the natural history parameters at a given `time`
+since the person was first infected. The `gi_cdf` describes the fraction of infectiousness that has occured
+in this parameter set a given time. Additional columns like `viral_load` and other time-varying quantities
+can easily be added in this input schema.
```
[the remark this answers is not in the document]
ChiragKumar9: Thanks. Could you help me understand why separate CSVs would be better here? I think Guido and I were against separate CSVs because that could mean that a user could update one CSV but just forget to update the other. We wanted to try to prevent that. Also, if we make an ixa helpers Python package, that could have a function that stitches together multiple incoming CSVs from the user into one main file to feed to ixa.

docs/natural-history-inputs.md
```diff
@@ -0,0 +1,95 @@
...
+the distribution of each of these parameters outside of Rust, but the user is also able to impose any
+meaningful correlations between parameters.
+
+## Data input format
+
+The input CSV includes samples of the natural history parameters. For now, we are interested in a single
+natural history parameter: the generation interval CDF over time. Data must be input in a long format, as
+such:
+
+| id | time | gi_cdf |
```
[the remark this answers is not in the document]
ChiragKumar9: Thanks. I meant this is long format in that each column doesn't isn't associated with a particular time and the value in that for a particular row is a sample at that time from the GI CDF like so:

| t=0 | t=0.02 | t=0.04 | t=0.08 | ... |
| --- | --- | --- | --- | --- |
| 0 | 0.001 | 0.002 | 0.003| ... |
| ... | ... | ... | ... | ... |

Comparatively, in the current input schema, I would add another column like so:
| id | time | gi_cdf | viral_load |
| --- | --- | --- | --- |

Am I correct in understanding that you are proposing a switch to an input schema like so?

| id | time | parameter | value |
| --- | --- | --- | --- |

docs/natural-history-inputs.md
```diff
@@ -0,0 +1,95 @@
...
+directly reproducing the variability present in these inputs in what is actually simulated.
+
+`mod natural_history_manager` provides a trait extension `ContextNaturalHistoryExt` that enables a user
+to query an individual's various natural history parameters during their infection. First, the user's
+natural history index must be set. We assume that it is set when an agent is first infected in
+`mod transmission_manager`. This implicitly means that all other modules that require knowing the agent's
+natural history parameter index (for instance, the future `mod health_status_manager` which requires
+knowing an agent's symptom onset and improvement times) must have their event callbacks for the
+`InfectiousStatusValue::Susceptible --> InfectiousStatusValue::Infectious` transition happen
+_after_ `mod transmission_manager`'s callback for the same transition.
```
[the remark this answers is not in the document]
ChiragKumar9: Ah, I struggled with the issue of assigning the index when drafting this PR, but I understand now that lazy assignment would fix this issue of one module relying on the other to set the index before it. Thanks!

docs/natural-history-inputs.md
```diff
@@ -0,0 +1,95 @@
+# A central natural history manager
+
+## Overview
+We provide a way for reading in user-specified natural history parameters (infectiousness over time/the
+generation interval, viral load over time, etc.) and parameterizing transmission based on these parameters.
```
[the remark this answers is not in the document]
ChiragKumar9: Oops, the Brits would have been happy here. I should have gone all out for and used an "s".

docs/natural-history-inputs.md
```diff
@@ -0,0 +1,95 @@
...
+different times. This could be useful if synthesizing various clinical datasets that each record the
+viral load at different times and a user wants to include the raw data in the model without doing
+any pre-processing, though it is worth noting that the module is set up to use cubic spline
+interpolation for intermediary values (more on this below.)
+
+## Implementation
+
+People are assigned an `id` when they are infected. This `id` refers to the natural history parameter
+set that will define their infection (for now, just the GI, but this will be expanded to include
+symptom onset and improvement times). `id`s are assigned uniformly and randomly, which allows for
```
confunguido, line 59: Are there any covariates for the sampling? For instance, I thought different symptoms were correlated with different type of curves, how are we handling that?

docs/natural-history-inputs.md
```diff
@@ -0,0 +1,95 @@
...
+directly reproducing the variability present in these inputs in what is actually simulated.
+
+`mod natural_history_manager` provides a trait extension `ContextNaturalHistoryExt` that enables a user
+to query an individual's various natural history parameters during their infection. First, the user's
```
confunguido, line 63: Who's the user? or why would the user have a natural history index? I'm confused.

docs/natural-history-inputs.md
```diff
@@ -0,0 +1,95 @@
...
+_after_ `mod transmission_manager`'s callback for the same transition.
+
+As of now, the `ContextNaturalHistoryExt` trait provides a method for querying the value of the
+inverse generation interval (i.e., returning a time from a uniform value) for scheduling
+[infection attempts](time-varying-infectiousness.md). The method takes the person doing the infecting
+as an input to query their natural history index and then use the particular generation interval CDF
+trajectory associated with that index. Since the generation interval samples that are provided are discrete,
+we use cubic spline interpolation (defaulting to linear interpolation at the tails) to estimate the value
+of the inverse CDF from the continuous uniform value provided as an input to the method. In the future,
+we may also add methods for querying the viral load, and they could use the same sort of interpolation
```
confunguido, line 78: This whole section is a bit difficult for me to follow.

confunguido: Avoid future statements that are not implemented here.

docs/natural-history-inputs.md
```diff
@@ -0,0 +1,95 @@
...
+2. There must be the same number of values for each parameter provided in the input CSV. In other words, a user
+cannot provide 1000 GI trajectories but only 10 symptom improvement times. The user must ensure that all parameter
+sets are complete and do that either via assuming independent draws between parameter values or imposing a correlation
+between parameter values. In other words, the burden of constructing reasonable parameter sets falls on the user
+rather than being implicitly baked into the model.
+3. We plan on introducing clinical symptoms in the same input CSV and being managed as another natural history
+parameter even though these outcomes are not necessarily only a function of the disease natural history. In other
+words, social determinants of health and other properties may also influence an agent's clinical outcomes from disease.
+Does this current input structure make sense, or is the inability to disentangle clinical outcomes from infectiousness
+dangerous? For our isolation guidance work, the two are inexplicably linked, and I cannot think of a way to disentangle
```
confunguido, line 94: This seems like a comment rather than a methods document.

confunguido (commented):
I provided initial review on markdown. I think there's plenty to improve there before going to the code.

docs/natural-history-inputs.md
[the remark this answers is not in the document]
ekr-cfa: Sorry, I was actually thinking about lines 38 below where the variables were constant in time

docs/natural-history-inputs.md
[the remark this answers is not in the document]
ekr-cfa: I'm not sure I'm proposing anything, but I had understood that "long" meant the last example you mentioned and that the middle one was kind of halfway between wide and long

docs/natural-history-inputs.md
```diff
@@ -0,0 +1,70 @@
+# A central natural history manager
+
+## Overview
+
+We provide an interface to read in user-specified natural history parameters, such as the generation
+interval distribution, and query the value of natural history parameters in modules that manage an
+individual's disease, like the transmission module.
+
+The goal of this module is to be flexible to modeling any natural history parameter that follows an
```
confunguido, line 9: Can you list such parameters.

src/natural_history_manager.rs
ChiragKumar9: Thanks. I initially used just linear interpolation and I noticed that I needed a much higher density of points than I anticipated, prompting the cubic interpolation. Basically:

1. Since CDFs can be either concave up or concave down at a given point, linear interpolation will always overestimate or underestimate the CDF, respectively. This results in consistent bias in the estimate of the inverse CDF, though the direction of that bias may change. By fitting to more points, cubic interpolation largely fixes this problem.
2. When doing linear interpolation, I had to provide CDF values down to 0.02 days in order to reproduce an exponential distribution with a KS stat of less than 0.01. That's a lot of points, and the model took a long time to start up. Comparatively, when using cubic interpolation, I could easily get away with CDF values down to 0.2 days, resulting in nearly a 5x speed up in just running the tests.
3. I don't think we should expect users to provide a given density of points. I agree that interpolation comes with a certain degree of "proceed with caution" but I also think the degree/type of interpolation used here is rather straightforward and standard.

I need to think more about doing the interpolation once -- I like the implications of this idea for the structure of the model, but my initial intuition is that you would end up not needing the interpolated values over a large range of the CDF in the actual simulation.

src/natural_history_manager.rs
ChiragKumar9: In this case, we're extrapolating the inverse CDF, so the $x$ values are the values of the generation interval CDF and the $y$ values are time. We can't extrapolate above 1, but you are correct that we can extrapolate some time that is very far away in the future as $xp \to 1$. But, that's also the behavior we expect from the tail of a CDF.

I realize now that I have an implicit assumption here: if the CDF were bounded -- meaning there were some maximum time value possible -- the input CSV would contain that maximum CDF value and time (i.e., a point that takes the form CDF value: 1.0000, time: 42.0). I think this is a fair assumption, but it's worth stating it. I'll add this to the markdown description as well. Thanks!

src/natural_history_manager.rs
ekr-cfa: I'll defer to @jasonasher about the math parts of it, but a few thoughts on the software engineering.

I don't think we should decide what to do here based on the test running time:

* Your current trajectory file is 4K lines. Reading 40K lines worth of CSV ought to be really fast, so if it's not, then I worry we have a different problem.
* If for some reason it's not fast, can we load it once into a fixture and cache it?

It seems like we're going to need to be doing interpolation a lot (for essentially each infection attempt, right?) so there is a lot of incentive to make that as fast as possible, even at considerable upfront cost, especially if that cost can be borne outside the simulation.

src/natural_history_manager.rs
```diff
@@ -0,0 +1,656 @@
...
+        }),
+    );
+}
+
+/// An interpolation routine that expects a paired set of values `xs` and `ys` that represent samples
+/// from a given function. The function is evaluated at a given x value `xp` using cubic spline interpolation
+/// when there are at least two samples above and below `xp`. Otherwise, it uses linear extrapolation at the tails.
+/// Assumes that function samples are sorted so that the `xs` are in ascending order.
+fn interpolate(xs: &[f64], ys: &[f64], xp: f64) -> f64 {
+    let window_right_edge_option = xs.iter().position(|&x| x > xp);
```
ekr-cfa, line 239: I believe that `.iter().position()` does a linear scan. I would recommend doing binary search.

Moreover, if I understand the algorithm correctly, for a given agent each query will have a larger value of `xp` than the previous query in which case we could remember that position, thus saving more time going forward.

ChiragKumar9:
Closing in favor of discussing the right API before implementing it, such as in #36.

PR #31, opened 2024-12-17
Intervention manager skeleton and facemask example

KOVALW:
> Currently out of date overview, see comment below and [documentation here](https://github.com/CDCgov/ixa-epi-isolation/blob/3e1f1567a595d23d4abe8fd596650387e99840ea/docs/infection.md). Keeping for record purposes
### Intervention skeleton
This PR introduces a generalized intervention manager that maps an intentionally ambiguous intervention type to an effect on relative transmission. The intervention manager has a register function and a query function to accomplish this. The `InterventionPlugin` uses a nested `HashMap` to relate infectious status to the intervention - relative transmission effect, and thus query can obtain both the relative risk of `Susceptible` individuals and the relative infectiousness of `Infectious` individuals. The register function therefore requires an `InfectiousStatus` to properly register the relative transmission, defaulting to 1.0 if not specified by the user.

### Other changes
Transmission success is now no longer guaranteed; success depends on the additive effects of relative risk and infectiousness of the contact and transmitter involved in the infection attempt. During infection attempts we take a uniform draw and consider the attempt a success if the draw is less than the aggregate relative transmission.

### Facemask intervention example
This manager is an example of defining an `enum` intervention type and registering it using the intervention manager. The facemask manager assigns each person a status of `Wearing`, at a rate defined in the parameters input, or `None`. Included in the tests of the facemask manager are some experiments to test that the epidemic size in a population of 500 is approximately what we would expect from the effective $R_0$ generated by the intrinsic rate of reproduction and the impact of facemasks.

### Next steps
- The intervention manager will need to be extended to allow the manager to query for infection attempts that are impacted by contact rate modifying interventions, not just infection success modifying interventions. This would preferably be done in a way that allows us to use the same query function, but accepting a boolean transmission attempt effect (contact or transmission success) that modifies a different function.
- The tests in the facemask example will need to be safeguarded against changes to the transmission manager, meaning that user-specified functions will need to allow for the current implementation.
- Facemask manager will likely need to read in a CSV of relative transmission effects rather than leaving them to be hard-coded.


KOVALW:
## Overview
Updated docs to reflect desired API (infection.md) along with skeleton code (transmission_modifiers.rs). All of the detailed description of implementation lives in the [infection.md](https://github.com/CDCgov/ixa-epi-isolation/blob/3e1f1567a595d23d4abe8fd596650387e99840ea/docs/infection.md). 

Facemasks as an instance of transmission modifiers can be put in once feedback is received on current skeleton, tests are made for transmission_modifiers.rs, and the hook to transmission_manager through `evaluate_transmission` is created. 

## Open concerns
- Are transmission modifier or intervention the terms we want to use? Current naming is a mix of `InterventionContainer` (with inner components of the data container following that naming convention, such as `intervention_map`) and `ContextTransmissionModifierExt` (with some functions and the documentation referring to relative transmission potential modifications). I would propose one of the following three terminologies be used, but I'm willing to hear other options
>1. Transmission Modifier (TransmissionModifierContainer, modifier_map, ContextTransmissionModifierExt, transmission_modifier.rs)
>2. Intervention Modifier (InterventionContainer, intervention_map, ContextInterventionExt, intervention_modifier.rs)
>3. Infection Attempt Manager (InfectionAttemptContainer, infection_attempt_map, ContextInfectionAttemptExt, infection_attempt_manager.rs)

- The `instance_map` currently throws a mismatched type error that doesn't allow a closure to be placed where the `intervention_map` expects an `InterventionFn`
- Related is that a default analogous to that supplied for the `aggregator` should be supplied for the `intervention_map` function. The implementation will have to be different due to the call in `compute_intervention`.

docs/infection.md
```diff
@@ -0,0 +1,163 @@
...
+
+```rust
+trait ContextTransmissionModifierExt {
+    fn register_intervention<T: PersonProperty + 'static + std::cmp::Eq + std::hash::Hash>(&mut self, infectious_status: InfectiousStatusType, person_property: T, instance_dict: Vec<(T::Value, f64)>);
+    fn register_aggregator(&mut self, agg_functions: Vec<(InfectiousStatusType, Box<AggregatorFn>)>);
+    fn compute_intervention(&mut self, person_id: PersonId) -> f64;
+}
+```
+
+To `register_intervention` for some instance of a modifier, the user must specify the `InfectiousStatus` and `PersonProperty` of interest and then provide an `instance_dict` that associates `PersonProperty` values to `float`s. These inputs are supplied to a default closure that obtains the realtive trasnmission potential modification through direct value association. We will want to abstract out this closure so that a generic function of type `InterventionFn` can be supplied instead.
```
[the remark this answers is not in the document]
ChiragKumar9: Ok, I see below that you say that `PersonProperty` values are connected to a given float. Again, I think you need to explain what these float/efficacy/modifier values are up front.

ChiragKumar9, line 89:
```suggestion
To `register_intervention` for some instance of a modifier, the user must specify the `InfectiousStatus` and `PersonProperty` of interest and then provide an `instance_dict` that associates `PersonProperty` values to `float`s. These inputs are supplied to a default closure that obtains the relative transmission potential modification through direct value association. We will want to abstract out this closure so that a generic function of type `InterventionFn` can be supplied instead.
```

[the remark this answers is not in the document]
ChiragKumar9: I think you should motivate _why_ the generic function is useful.

docs/infection.md
```diff
@@ -0,0 +1,163 @@
...
+                return item.1;
+            }
+        }
+        // Return a default 1.0 (no relative change if unregistered)
+        return 1.0;
+    });
+
+    let intervention_container = self.get_data_container_mut(InterventionPlugin);
+
+    // mismatched types: instance_map is of type {closure} but expected {dyn Fn(...)}
```
ChiragKumar9, line 109: ?

docs/infection.md
```diff
@@ -0,0 +1,163 @@
...
+    intervention_container
+    .intervention_map
+    .insert(infectious_status, instance_map);
+
+}
+```
+
+To register the `aggregator` functions, because we only can only have as many aggregators as there are `InfectiousStatusType` values, the user provides a `Vec` of all aggregator functions desired to the register function.
```
ChiragKumar9, line 117: I don't understand how the requirement that we have as many aggregators as infectious status types means we should provide a vector?

docs/infection.md
```diff
@@ -0,0 +1,163 @@
...
+        intervention_container
+        .aggregator
+        .insert(item.0, item.1);
+    }
+}
+```
+
+We already defined the default aggregator through `impl InterventionContainer` so that the user does not need to specify all or even any aggregators in order to `run_aggregator` in `compute_intervention`.
+
+To calculate the total relative trasnmission potential change for an individual during an infection attempt, `Context` and the `PersonId` are supplied to `compute_intervention`, which is agnostic of the `InfectiousStatus` of the `PersonId` supplied.
```
ChiragKumar9, line 133:
```suggestion
To calculate the total relative transmission potential change for an individual during an infection attempt, `Context` and the `PersonId` are supplied to `compute_intervention`, which is agnostic of the `InfectiousStatus` of the `PersonId` supplied.
```

Are we calculating a "change" in relative transmission potential? I also don't think the function is agnostic to the infectious status -- it just knows to query the infectious status from the person's ID.

docs/infection.md
```diff
@@ -0,0 +1,163 @@
...
+        .unwrap();
+
+    for (t, f) in intervention_map {
+        registered_interventions.push((*t, f(self, person_id)));
+    }
+
+    intervention_plugin.run_aggregator(infectious_status, &registered_interventions)
+}
+```
+We use automatic detection of all the modifications applied to a person, which is necessary to remove multiple manual calls to query particular interventions that would otherwise be error-prone and inflexible.
```
ChiragKumar9, line 153: I don't think we automatically detect the modifications applied to a person. We just iterate through all available interventions that have been registered and see if any of them are applicable to the person. If there is an intervention that hasn't been registered, we don't "automatically" detect it.

I also think the point made in this sentence is really about creating a structure that encourages model modularity.

docs/infection.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Querying
+We provide the simple public function to query all the relative transmission modifiers for both S and I during the infection attempt
+
+```rust
+
+pub fn query_modifers(context: &mut Context, transmitter_id: PersonId, contact_id: PersonId) -> f64 {
```
ChiragKumar9, line 160: Same comment about this being part of the trait extension.

ChiragKumar9 (changes requested):
Partial review on the design doc. Will review code tomorrow.

ekr-cfa (changes requested):
At a high level, I don't think having a big design that describes the implementation separate from the code is that helpful. This is reasonably straightforward code and so inline comments should be sufficient. To the extent to which they are not, I would prefer whole file docs, either as API doc or just at the top of the file. That keeps it with the code.

If you want to have a document which describes the model structure rather than the code, I think that's reasonable.

src/transmission_modifiers.rs
```diff
@@ -0,0 +1,121 @@
...
+        &mut self,
+        infectious_status: InfectiousStatusType,
+        person_property: T,
+        instance_dict: Vec<(T::Value, f64)>,
+    );
+    fn register_aggregator(
+        &mut self,
+        agg_functions: Vec<(InfectiousStatusType, Box<AggregatorFn>)>,
+    );
+    fn compute_intervention(&mut self, person_id: PersonId) -> f64;
```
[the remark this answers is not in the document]
KOVALW: The effect across interventions - is `compute_relative_transmission clearer`?

src/transmission_modifiers.rs
```diff
@@ -0,0 +1,121 @@
...
+        instance_map.insert(
+            TypeId::of::<T>(),
+            move |context: &mut Context, person_id| -> f64 {
+                let property_val = context.get_person_property(person_id, person_property);
+
+                for item in instance_dict {
+                    if property_val == item.0 {
+                        return item.1;
+                    }
+                }
```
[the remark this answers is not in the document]
KOVALW: I don't think that we would have more than a few. Robert's comments made it seem like several hundred would be the turning point for preferring Hash. I don't think we would ever have a case where something like PersonId is a key

src/transmission_modifiers.rs
```diff
@@ -0,0 +1,121 @@
...
+    fn register_intervention<T: PersonProperty + 'static + std::cmp::Eq + std::hash::Hash>(
+        &mut self,
+        infectious_status: InfectiousStatusType,
+        person_property: T,
+        instance_dict: Vec<(T::Value, f64)>,
+    );
+    fn register_aggregator(
+        &mut self,
+        agg_functions: Vec<(InfectiousStatusType, Box<AggregatorFn>)>,
+    );
```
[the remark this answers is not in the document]
KOVALW: I like the option of having to specify each over all. Agreed that the difference between registering modifiers and aggregators was confusing. Changed to be a single call per InfectiousStatusType

docs/transmission_modifiers.md
```diff
@@ -0,0 +1,297 @@
...
+                aggregate_effects *= effect;
+            }
+
+            aggregate_effects
+        })
+    }
+}
+```
+
+We define a default aggregator so that the user does not need to specify all or even any
```
ChiragKumar9, line 205: One design caveat I see here is that the aggregator function registration is not modular. If a user wants to register their own custom modifier function, they need to be aware of all other interventions and their `TypeId`s. In the short term, we should write this down as a caveat. In the long term, I wonder if we can design an interface so that you can specify how two (or some subset of all interventions) interact with a custom aggregator and then the rest of the interventions default to using the independence aggregator.

docs/transmission_modifiers.md
```diff
@@ -0,0 +1,297 @@
...
+context.register_transmission_modifier(
+    infectious_status: InfectiousStatusType::Infectious,
+    person_property: FacemaskStatus,
+    modifier_key: &vec![(FacemaskStatusType::Wearing, 1.0 - parameters.facemask_infectious_efficacy)],
+    )
+```
+
+### Hook to aggregator from PersonId
+To calculate the total relative transmission potential change for an individual during an infection
+attempt, `Context` and the `PersonId` are supplied to `compute_relative_transmission`, which
```
ChiragKumar9, line 259: Nit: `Context` isn't provided as an argument. The method is implemented as part of a trait extension on `Context`.

docs/transmission_modifiers.md
```diff
@@ -0,0 +1,297 @@
...
+
+    for (t, f) in transmission_modifier_map {
+        registered_modifiers.push((*t, f(self, person_id)));
+    }
+
+    transmission_modifier_plugin.run_aggregator(infectious_status, &registered_modifiers)
+}
+```
+
+We use automatic iteration through all the modifications applied to a person, which is necessary to
```
ChiragKumar9, line 281:
```suggestion
We automatically apply each transmission modifier function to determine all the modifications applied to a person, which is necessary to
```

src/transmission_modifiers.rs
ChiragKumar9, line 1: What do you think about naming this file `transmission_modifiers_manager.rs`? `transmission_modifiers` on its own sounds like it is a list of transmission modifiers, not that it manages those modifiers.

ChiragKumar9 (commented):
I think the document could use one more clean up to better articulate that we are modifying relative probability of transmission. I put a bunch of changes here to help guide it in that direction, but I acknowledge that part of this is taste and how we are each independently choosing to conceptualize the idea between an infection attempt, reduction in infection, etc.

docs/transmission_modifiers.md
```diff
@@ -0,0 +1,297 @@
...
+`query_infection_modifiers` function of the context trait extension
+`ContextTransmissionModifierExt`.
+
+```rust
+fn query_infection_modifers(&mut self, transmitter_id: PersonId, contact_id: PersonId) -> f64 {
+    self.compute_relative_transmission(transmitter_id) * self.compute_relative_transmission(contact_id)
+}
+```
+
+We assume that the infectiousness of transmitter I is independent of, and therefore additive
```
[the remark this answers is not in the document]
KOVALW: For me additive = independent, but that's just my tendency to think in log-space

docs/transmission_modifiers.md
```diff
@@ -0,0 +1,297 @@
...
+Other modules will implement particular instances of transmission modifiers and the
+`transmission_manager` will call `query_infection_modifiers`, meaning that these modules will
+require access to the `TransmissionModifierContainer`. We therefore require a trait extension that
+includes the functionality to register modifiers and their aggregators, as well as call an
+aggregator to compute the total relative transmission potential change due to the modifications
+present for a given `PersonId`.
+
+```rust
+trait ContextTransmissionModifierExt {
+    fn register_transmission_modifier<T: PersonProperty + 'static + std::cmp::Eq + std::hash::Hash>(&mut self, infectious_status: InfectiousStatusType, person_property: T, instance_dict: Vec<(T::Value, f64)>);
```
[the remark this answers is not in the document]
KOVALW: These are actually unnecessary calls and should be removed (per Eric's comment [above](https://github.com/CDCgov/ixa-epi-isolation/pull/31#discussion_r1906446190)). They're still in here but not in the script

confunguido (changes requested):
I only reviewed the markdown document. See my comments. I am not entirely sure how this module should, if at all, connect with what we call now natural history module. For instance, how would anti-virals change transmissibility, is this at the instantaneous transmission attempt or does it modify the infectiousness over time that will result in changes in the timing of infection attempts?

Issue #28, opened 2024-12-10
Idiomatic ixa


ChiragKumar9:
## Goal
Update the use of ixa in this repo to match the latest and greatest syntax on the the ixa repo's main branch.

## Context
Ixa development has been ongoing while we have been using it in this repo. As such, we may have syntax that still works but is no longer the best way to do something in our repo. We should update to use the desired ixa syntax.

## Specifications
- [ ] Use trait extensions exported from main ixa crate rather than accessing them from a module within ixa.
- [ ] Use ixa's built-in method `sample_person` to get a random person at the beginning of the simulation to infect.
- [ ] Once merged in, switch to using ixa's built-in command line runner.

k88hudson-cfa:
Runner has landed: https://github.com/CDCgov/ixa/pull/114

PR #32, opened 2024-12-18
Update ixa - Closes issue #28

confunguido:
As described in issue #28, removed unnecessary module name in ixa imports, updated main to use ixa runner with custom args, and use `sample_person`. A couple of things to highlight:
1) sample_person does not yet allow to filter by a person property, so making a safe assumption here that everyone is alive at time t = 0.
2) It's a bit confusing to use custom_args for boolean values. For instance, I used the `force_overwrite` flag to determine whether or not to overwrite reports. If not specified, I would assume the value is false, but if not specified, `custom_args` would not exist, so I still need to do `if let Some(custom_args) = custom_args {}else{}`, which seems a bit unnecessasry.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -17,14 +17,13 @@ use crate::population_loader::{Age, CensusTract};
 
 #[derive(Args, Debug)]
 struct CustomArgs {
-    /// whether force overwrite of output files if they already exist
+    ///Whether force overwrite of output files if they already exist
     #[arg(short = 'f', long)]
     force_overwrite: bool,
 }
 
 fn initialize() -> Result<Context, IxaError> {
     let mut context = run_with_custom_args(|context, args, custom_args: Option<CustomArgs>| {
-        println!("Setting Overwrite parameter");
         // Read the global properties.
         let custom_args = custom_args.unwrap();
         // Set the output directory and whether to overwrite the existing file.
--- a/src/main.rs
+++ b/src/main.rs
@@ -3,10 +3,11 @@ mod parameters;
 mod population_loader;
 mod transmission_manager;
 
-use clap::Parser;
+use ixa::runner::run_with_custom_args;
+use clap::Args;
 use ixa::{
-    context::Context, error::IxaError, global_properties::ContextGlobalPropertiesExt,
-    people::ContextPeopleExt, random::ContextRandomExt, report::ContextReportExt,
+    Context, IxaError, ContextGlobalPropertiesExt,
+    ContextPeopleExt, ContextRandomExt, ContextReportExt,
 };
 use std::path::PathBuf;
 use transmission_manager::InfectiousStatus;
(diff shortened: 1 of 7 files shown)


confunguido:
I can see that my tests aren't working, probably pre-commit issue. Working on fixing that.

ekr-cfa:
I just added `sample_person()` with a query

confunguido:
> I just added `sample_person()` with a query

Thanks! Just updated the code.

confunguido:
> As described in issue #28, removed unnecessary module name in ixa imports, updated main to use ixa runner with custom args, and use `sample_person`. A couple of things to highlight:
> 
> 1. sample_person does not yet allow to filter by a person property, so making a safe assumption here that everyone is alive at time t = 0.
> 2. It's a bit confusing to use custom_args for boolean values. For instance, I used the `force_overwrite` flag to determine whether or not to overwrite reports. If not specified, I would assume the value is false, but if not specified, `custom_args` would not exist, so I still need to do `if let Some(custom_args) = custom_args {}else{}`, which seems a bit unnecessasry.

All my initial comments about `sample_person` and custom boolean args have been solved with ixa updates.

src/main.rs
```diff
@@ -3,50 +3,47 @@ mod parameters;
...
-    #[arg(short, long)]
-    input_file: PathBuf,
-
-    /// path to the output directory
-    #[arg(short, long)]
-    output_directory: PathBuf,
-
+#[derive(Args, Debug)]
+struct CustomArgs {
     /// whether force overwrite of output files if they already exist
```
ChiragKumar9, line 20:
```suggestion
    /// Whether to force overwrite of the output files if they already exist.
```

src/main.rs
```diff
@@ -3,50 +3,47 @@ mod parameters;
...
     force_overwrite: bool,
 }
 
-fn initialize(args: &Args) -> Result<Context, IxaError> {
-    let mut context = Context::new();
-    // Read the global properties.
-    context.load_global_properties(&args.input_file)?;
+fn initialize() -> Result<Context, IxaError> {
+    let mut context = run_with_custom_args(|context, args, custom_args: Option<CustomArgs>| {
+        println!("Setting Overwrite parameter");
```
ChiragKumar9, line 27: I assume this was for debugging purposes. I would remove this print statement now and not have it on main.

src/transmission_manager.rs
```diff
@@ -1,15 +1,13 @@
...
-    global_properties::ContextGlobalPropertiesExt,
-    people::{ContextPeopleExt, PersonId, PersonPropertyChangeEvent},
-    random::ContextRandomExt,
+    define_person_property, define_person_property_with_default, define_rng, Context,
+    ContextGlobalPropertiesExt, ContextPeopleExt, ContextRandomExt, IxaError, PersonId,
+    PersonPropertyChangeEvent,
 };
 use statrs::distribution::{ContinuousCDF, Exp, Poisson};
 
+use crate::contact::ContextContactExt;
```
ChiragKumar9, line 8: Nit: possible to make one `use crate::{}` import statement? I find that easier to read.

ChiragKumar9 (changes requested):
Please also update the readme with the new syntax for running the model.

src/main.rs
```diff
@@ -17,14 +17,13 @@ use crate::population_loader::{Age, CensusTract};
 
 #[derive(Args, Debug)]
 struct CustomArgs {
-    /// whether force overwrite of output files if they already exist
+    ///Whether force overwrite of output files if they already exist
```
ChiragKumar9, line 20:
```suggestion
    /// Whether to force overwrite of output files if they already exist.
```

ChiragKumar9 (approved):

src/main.rs
```diff
@@ -3,50 +3,46 @@ mod parameters;
...
-    /// path to the output directory
-    #[arg(short, long)]
-    output_directory: PathBuf,
-
-    /// whether force overwrite of output files if they already exist
-    #[arg(short, long, default_value = "false", default_missing_value = "true")]
+#[derive(Args, Debug)]
+struct CustomArgs {
+    ///Whether force overwrite of output files if they already exist
+    #[arg(short = 'f', long)]
```
k88hudson-cfa, line 21: This will be `f` by default (you can just include `short`)

k88hudson-cfa (approved):
LGTM

PR #34, opened 2024-12-23
Gce ixa runner nocustom

confunguido:
Removed the need to use custom arguments to use force overwrite

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -3,77 +3,60 @@ mod parameters;
 mod population_loader;
 mod transmission_manager;
 
-use clap::Args;
-use ixa::runner::run_with_custom_args;
+use ixa::runner::run_with_args;
 use ixa::{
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -3,77 +3,56 @@ mod parameters;
...
+    run_with_args(|context, args, _| {
         // Read the global properties.
-        let custom_args = custom_args.unwrap();
         // Set the output directory and whether to overwrite the existing file.
         context
             .report_options()
-            .directory(PathBuf::from(&args.output_dir))
-            .overwrite(custom_args.force_overwrite);
+            .directory(args.output_dir.unwrap())
+            .overwrite(args.force_overwrite);
```
k88hudson-cfa, line 20: You can actually remove this entirely, the options get set automatically

k88hudson-cfa (approved):
You can remove the report option setting stuff entirely, looks good aside from that

ChiragKumar9 (approved):
Agree with Kate's suggestion.

PR #36, opened 2024-12-30
Design document: natural history manager

ChiragKumar9:
This PR proposes an API for a natural history manager: a module that stores any natural history parameter specified by the user and provides a way for other modules to query values of these parameters in a way that gives them the values necessary for modeling disease transmission at the individual level. Here's a short proof-of-concept in Rust playground for some of the implementation I introduce: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0fb34c6a4c9bcc3b05694435f10fed3a.

### DIFF
--- a/docs/natural_history_manager.md
+++ b/docs/natural_history_manager.md
@@ -21,9 +21,10 @@ individual-level. This manager also provides a trait extension on `Context` call
 `ContextNaturalHistoryExt` that contains convenience methods for common multi-step calculations on
 natural history parameters. The custom types defined in this module are all deserializable, so the
 user can specify natural history parameters in an input JSON file, taking advantage of the existing
-global properties plugin's infrastructure. We focus on having these types be as generic as possible,
-so that the user can change the value (for instance, changing an exponential distribution to a gamma
-distribution) without impacting the parameter type or the way it is used in the code.
+global properties plugin's infrastructure (in particular, parameter validation at read-in time). We
+focus on having these types be as generic as possible, so that the user can change the value (for
+instance, changing an exponential distribution to a gamma distribution) without impacting the
+parameter type or the way it is used in the code.
 
 ### Time-invariant parameters: An example from specifying the disease incubation period
 
@@ -39,7 +40,7 @@ pub enum ProbabilityDistribution {
     Exp(f64),
     Gamma(f64, f64),
     Empirical(Vec<f64>),
-    // ... Because this is an enum, users can add any other distribution variants.
+    // ... Because this is an enum, users can add other distribution variants.
 }
 ```
 
@@ -53,20 +54,21 @@ A user may specify time-invariant natural history parameters in the input JSON a
     "epi_isolation.Parameters": {
         "incubation_period": "Gamma(1.0, 3.0)",
         "time_to_symptom_improvement": "Empirical([1.0, 2.0])",
-        "R_0": "Empirical([3.0])"
+        "R_0": "Empirical([3])"
     }
 }
 ```
 
 In the case of the distribution for $R_0$, we are saying that all agents have the same value of
 $R_0$ -- there is no stochasticity or person-to-person variation in $R_0$. Nevertheless, because we
-have defined the `Distribution` type broadly, we can consider this special case. For our
-`ProbabilityDistribution` type, we implement Rust's `rand` crate's `Distribution` trait. This
-enables us to draw random samples from the distribution using Ixa's built-in `ContextRandomExt` that
-expects a type that implements `Distribution<T>` for sampling. This lets us treat our custom enum
-like we natively defined the distribution type in Rust without going through our enum first. To
-reduce confusion, we adopt that the meaning of parameters in our custom `ProbabilityDistribution`
-type is the same as in the `statrs` distribution of the same name.
+have defined the `ProbabilityDistributionDiscrete` type broadly, we can consider this special case
+within our defined parameter types. For our `ProbabilityDistribution` type, we implement Rust's
+`rand` crate's `Distribution` trait. This enables us to draw random samples from the distribution
+using Ixa's built-in `ContextRandomExt` that expects a type that implements `Distribution<T>` for
+sampling. This lets us treat our custom enum like we natively defined the distribution type in Rust
+without going through our enum first. To reduce confusion, we adopt that the meaning of parameters
+in our custom `ProbabilityDistribution` type is the same as in the `statrs` distribution of the same
+name.
 
 ```rs:natural_history_manager.rs
 use rand::distribution::Distribution
@@ -84,8 +86,7 @@ impl Distribution<f64> for ProbabilityDistribution {
 ```
 
 The user can also implement their own traits on the `ProbabilityDistribution` type if they needed
-to extract the raw `statrs` type (i.e., what they would have gotten if they called `Exp::new(1.0)`,
-for instance) or to do other computations like getting the mean or variance of the distribution.
+to do other computations like getting the mean or variance of the distribution.
 
 To draw a random sample in a module with this type, say to draw the time at which to set the
 person's health status to symptomatic once the agent becomes infected based on the incubation period
@@ -122,11 +123,11 @@ fn handle_infectious_status_change(
 The modular structure lends itself to being able to easily change the underlying distribution type
 (i.e., switching out a gamma for, say, a lognormal) in the input JSON with no change to the code and
 only needing to make sure the new distribution can be coerced into a Rust type that implements the
-`Distribution` trait. The incubation period parameters could even be updated in the middle of the
-simulation! Our `ProbabilityDistribution` type handles many use cases for natural history parameters
-and more broadly being able to specify and sample from arbitrary distributions in our ABMs. This
-type also serves as the building block for the next natural history parameter type we define:
-time-varying parameters.
+`Distribution` trait. The incubation period distribution could even be updated in the middle of the
+simulation by just changing the global properties! Our `ProbabilityDistribution` type handles many
+use cases for natural history parameters and more broadly being able to specify and sample from
+arbitrary distributions in our ABMs. This type also serves as the building block for the next
+natural history parameter type we define: time-varying parameters.
 
 ### Time-varying parameters: An example from specifying the respiratory viral load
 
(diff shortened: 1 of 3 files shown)


docs/natural_history_manager.md
```diff
@@ -0,0 +1,274 @@
...
+manager, the transmission manager's syntax is focused entirely on the person-to-person transmission
+workflow, independent of the mathematical calculations required to obtain the times. The
+transmission manager no longer has to store `last_infection_time_unif` -- the quantile of the last
+infection attempt time in the generation interval distribution, which is an uninterpretable
+quantity -- or separately manage the number of infection attempts and the time to an infection
+attempt.
+
+### Querying clinical-relevant natural history parameters
+
+Most ABMs are calibrated to counts from clinical data, such as the number of people who are
```
confunguido, line 150: I don't think we need that much context on this. Try to keep it simple.

docs/natural_history_manager.md
```diff
@@ -0,0 +1,274 @@
...
+input CSV.
+
+This input structure has two implications for time-varying parameters. First, there may be multiple
+parameters that vary over time, but they do not need to have values at the same time (for instance,
+the viral load and generation interval CDF for the same `index` could have samples at different
+time). Second, a time-varying parameter may have different time values across different `index`s.
+We use linear interpolation to estimate the value of a time-varying parameter at a continuous time
+value in the model from the samples provided at discrete times in the input CSV.
+
+## Application to isolation guidance
```
confunguido, line 253: I would move this higher in the document.

docs/natural_history_manager.md
```diff
@@ -0,0 +1,274 @@
...
+agent's natural history parameters. For the `time_to_next_infection_attempt` method, the relevant
+parameters are the number of secondary infection attempts and generation interval distribution (or
+an index that can be used to query the generation interval and any other natural history parameters
+from an input CSV; more on this below). We set these parameters as person properties in the
+`assign_natural_history(&mut Context, PersonId)` function.
+
+```rust
+pub trait ContextNaturalHistoryExt {
+    fn time_to_next_infection_attempt(&mut self, person_id: PersonId) -> Option<f64> {
+        assign_natural_history(self, person_id);
```
confunguido, line 58: I don't think we should assign natural history like this. Shouldn't these be two different processes? For instance, once infected, pick a natural history curve, then when scheduling infections look into this method. Question is how to connect this to transmission.rs

docs/natural_history_manager.md
```diff
@@ -0,0 +1,274 @@
...
+            return None;
+        }
+        // Calculate the next infection attempt time.
+        // First, get the last infection attempt times.
+        let (last_infection_attempt_unif, last_infection_attempt_time) = self.get_person_property(person_id,
+                                                                                                  LastInfectionAttemptTime);
+        // Order statistics math to get the next infection attempt time.
+        let next_infection_attempt_unif = get_next_infection_attempt_unif(infection_attempts_remaining,
+                                                                          last_infection_attempt_unif);
+        // Convert from uniform space to real time.
```
confunguido, line 70: Given that this is markdown, code comments seem redundant

docs/natural_history_manager.md
confunguido, line 1: As I read this document, the flow of the processes isn't clear.  I would appreciate an algorithm description that mentions for instance when the natural history is initialized, when an infection attempt is made, etc.

docs/natural_history_manager.md
```diff
@@ -0,0 +1,274 @@
...
+```
+
+The transmission manager would use the method as follows:
+
+```rust
+fn handle_infectious_status_change(
+    context: &mut Context,
+    event: PersonPropertyChangeEvent<InfectiousStatus>,
+) {
+    // Is the person going from S --> I?
```
confunguido, line 100: These comments are unnecessary, and wrong.

docs/natural_history_manager.md
```diff
@@ -0,0 +1,274 @@
...
+    // We don't care about the other transitions here, but this function will still be triggered
+    // because it watches for any change in InfectiousStatusType.
+    if event.current == InfectiousStatusType::Infectious {
```
confunguido, line 103: See my other comment about whether these should be two separate processes: assign parameters and schedule next infection attempt.

docs/natural_history_manager.md
```diff
@@ -0,0 +1,274 @@
...
+natural history parameters (or an index that can be used to query them as mentioned above), and
+each natural history querying method should call it so to ensure that all natural history parameters
+are set. If, for example, the `time_to_symptom_onset` method below does not call the
+`assign_natural_history` method, that would require that the natural history had already been
+assigned and that the `time_to_next_infection_attempt` method had been called prior to the
+`time_to_symptom_onset` method. This would require that the transmission manager's event listeners
+were triggered prior to the clinical symptoms event manager's. That structure introduces subtle
+dependencies between modules, making them loss modular and prone to bugs.
+
+To ensure that an agent's natural history parameters are not being changed in the middle of their
```
confunguido, line 172: This kinda makes me think that they should be separate processes. They do different things.

confunguido (changes requested):

docs/natural_history_manager.md
```diff
@@ -0,0 +1,274 @@
...
+transmission workflow.
+
+Determining the time to the next infection attempt requires the following:
+1. Knowing the number of infection attempts remaining for this agent.
+2. Knowing the timing of their last infection attempt.
+3. Knowing the agent's disease generation interval distribution.
+4. Using order statistics to calculate the time to the next infection attempt based on the above
+quantities.
+
+When the `time_to_next_infection_attempt` method is called for the first time, it needs to set the
```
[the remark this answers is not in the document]
ChiragKumar9: We discussed that the index assignment should be lazy, and it makes sense to group it into getting the parameter value because the two must always be called together.

docs/natural_history_manager.md
```diff
@@ -0,0 +1,274 @@
...
+    // because it watches for any change in InfectiousStatusType.
+    if event.current == InfectiousStatusType::Infectious {
+        schedule_next_infection_attempt(
+            context,
+            event.person_id,
+        );
+    }
+}
+
+fn schedule_next_infection_attempt(
```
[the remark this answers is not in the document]
ChiragKumar9: done

docs/natural_history_manager.md
ChiragKumar9: I believe what you are suggesting favors an explicit assignment of natural history parameters. I don't think that's necessary as it just pushes more code (and the potential to forget it) to the calling module for `get_parameter_value()`. Happy to discuss if I am misunderstanding.

docs/natural_history_manager.md
```diff
@@ -0,0 +1,512 @@
...
+estimation, whether we want to actually be estimating the _inverse_ of the function (in this case,
+that is `true`), and a starting point. The starting point argument is useful because we often need
+to get the value of a time-varying parameter over the course of an infection. Time since being
+infected only ever increases, so we know that we will be querying the function at larger times.
+Depending on the underlying type for which the trait is being implemented, knowing the starting
+point for doing our estimation routine can help improve computational efficiency. We will explain
+the details of implementing this trait below, but the abstractness of being able to tailor the
+implementation of the trait to the type helps improve modularity.
+
+### A general pattern for getting parameter values
```
KOVALW, line 216: This is a sufficiently simpler explanation of how `NaturalHistoryParameterValue` can be either scalar or vector and the treatment of each that it might be worthwhile to bring this up earlier to prime the reader for following `time_to_next_infection_attempt`

docs/natural_history_manager.md
```diff
@@ -0,0 +1,512 @@
...
+`NaturalHistoryParameterValue` is object safe, so we can store parameter values in a vector and
+store that vector in a hash map entry referring to its parameter value:
+`HashMap<NaturalHistoryParameter, Vec<Box<dyn NaturalHistoryParameterValue>>>`. The
+`get_parameter_value()` method just queries this hash map, returning the value of the specified
+natural history parameter for the specified person.
+
+Although `get_parameter_value()` takes the person ID as an input, how does the method know which
+parameter value should be returned for the person in question? We adopt the convention that each
+person is assigned an index, the `NaturalHistoryIndex`. This index is used when taking a value from
+the vector in the hash map for a specified natural history parameter. Indeces are assigned based on
```
KOVALW, line 365: indices or indexes (the latter works better as multiple assignments of an individual label), not indeces

docs/natural_history_manager.md
```diff
@@ -0,0 +1,512 @@
...
+weights. Together, this tells us the two features of the natural history data container that stores
+the parameter values:
+
+```rs:natural_history_manager.rs
+#[derive(Default)]
+struct NaturalHistoryContainer {
+    weights: Vec<f64>,
```
KOVALW, line 372: I'm concerned that weights across sets of `NaturalHistoryParameterValues` would restrict the case of user-defined distributions for calling the parameters. Would these still need to be defined in the HashMap even though they would be global values (or at least group values) for defining the rulesets for individual draws during runtime? I'm thinking that if that's the case you waste a lot of CSV input rows on repeat values.

docs/natural_history_manager.md
```diff
@@ -0,0 +1,512 @@
...
+knows about. Parameters are specified by an index, the same index that an agent is assigned
+(proportional to the specified weight) when their natural history parameters are queried. Grouping
+parameters by indeces allows for implicit correlations and dependencies between parameters. For
+instance, in this example `index = 0` describes the infection of an individual who has a symptomatic
+infection because they have an incubation period and time to symptom improvement. Both are required
+to specify a symptomatic infection. Likewise, the generation interval CDF for this person can be
+specific to them being a symptomatic individual. Implicitly, all values of a parameter for a given
+index represent a joint sample from the data generation process for the natural history of
+infectious agents. On the other hand, the agent who is assigned `index = 1` has an asymptomatic
+infection because these parameters are not present in their parameter set. Therefore, this input
```
KOVALW, line 475: Information by omission seems like a great way to have hidden bugs in the preprocessing steps. I'd rather see a value that ixa-epi-isolation interprets as asymptomatic (None) so that we know the inputs were correctly generated

KOVALW (changes requested):
This is much improved over having this functionality in the transmission manager and I like the schema in general. There are a couple of changes to consider, particularly with reading in the parameter values.

PR #39, opened 2025-01-08
This should enable Rust support in codespaces.

ekr-cfa:


ChiragKumar9 (commented):
Thanks. I'm rather naive about codespaces, so I may not be the best reviewer here. I could get `cargo build` and `cargo run` to work, but I had to install rustup first. Is this expected behavior?

ChiragKumar9:
@k88hudson-cfa I still need to manually run the install even with these changes. Is it worth including an install script like here? https://github.com/codespaces-examples/rust/blob/main/.devcontainer/setup.sh Or I could just be missing something about how to set up codespaces.

k88hudson-cfa (approved):

PR #40, opened 2025-01-24
Proposing an API for specifying the natural history parameters

ChiragKumar9:
Revamped version of #36 inspired by Jason and Guido's suggestions to think about toggling between transmission model "modes" and how we can make specifying simple disease models and complex custom models equally easy.

Once we have agreement on the API, I am happy to make small PRs with sketches of the code that implements the ideas presented here, but I want us to have agreement on the API first.

### DIFF
--- a/docs/0_contents.md
+++ b/docs/0_contents.md
@@ -3,10 +3,11 @@
 1. Synthetic population
 2. [Transmission](transmission.md)
     - [Time-varying infectiousness](time-varying-infectiousness.md)
-    - Estimating infectiousness from viral load
+    - [Storing and querying disease natural history](natural_history_manager.md)
 3. Contact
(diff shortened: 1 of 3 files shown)


k88hudson-cfa:
A few pieces here that I am hoping you can implement here using `InfectiousnessRateFunction` in #48, specifically:
* Loading viral load trajectories so they can be assigned to individuals
* An implementation of `rate`, `cum_rate`, `inverse_cum_rate` for emprical distributions using interpolation as described here, and based on your previous work

One part I'm not exactly sure about is what you mean about correlated parameter values / how that might be expressed – happy to discuss more as that comes up :)

confunguido (commented):
I will push some of these changes for completion that I had in another branch.

PR #43, opened 2025-01-25
Update Github actions precommit workflow

ChiragKumar9:
Not sure why CI in #41 didn't complain that the pre-commit workflow needed packages cached. When I include this commit in #40, pre-commit on Github Actions passes. While I was at it, I updated the styler and used ruff over black in line with the current CFA repo template.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -54,3 +54,11 @@ repos:
     -   id: detect-secrets
         args: ['--baseline', '.secrets.baseline']
         exclude: package.lock.json
+-   repo: https://github.com/doublify/pre-commit-rust
+    rev: v1.0
+    hooks:
+    -   id: fmt
+        args: ["--verbose", "--"]
+    -   id: clippy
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -59,11 +54,3 @@ repos:
     -   id: detect-secrets
         args: ['--baseline', '.secrets.baseline']
         exclude: package.lock.json
--   repo: https://github.com/doublify/pre-commit-rust
-    rev: v1.0
-    hooks:
-    -   id: fmt
-        args: ["--verbose", "--"]
-    -   id: clippy
```
KOVALW, line 67: Does the ruff fmt still allow for clippy warnings to throw pre-commit errors? This seems like it's just removing the rust guardrails but I'm probably missing something.

KOVALW (commented):
Making sure that clippy and rust formatting is still followed, the cahcing looks good

confunguido (approved):

PR #51, opened 2025-02-08
Add new infectiousness loop, rate functions

k88hudson-cfa:
This PR:
* Implements a new infection propagation loop with forecast and evaluation functions, replacing the existing transmission manager.
* Introduces an `InfectiousnessRateFn` trait and related utilities for defining rate functions. It includes an initial implementation of `ConstantRate`, but is designed to be expanded to handle other kinds of distributions.

`InfectiousnessRateFn` is designed to support scaling by a multiplier or other rate function, as well as offsetting by some amount (i.e., the elapsed time of infection so far), which requires different implementations depending on the underlying rate function.

(Note, originally this implemented a basic `Settings` concept which allows you to get members and choose a next contact, but I removed it in review to be addressed in a follow-up after some design work)


src/infection_propagation_loop.rs
```diff
@@ -0,0 +1,247 @@
...
+        seed_infections(context, initial_infections);
+    });
+
+    // Add a plan to shut down the simulation after `max_time`, regardless of
+    // what else is happening in the model.
+    context.add_plan(max_time, |context| {
+        context.shutdown();
+    });
+
+    context.subscribe_to_event::<PersonPropertyChangeEvent<InfectiousStatus>>(|context, event| {
```
confunguido, line 98: It's a bit confusing that the type is InfectiousStatus, but the value is Infected. Should it be InfectionStatus? or the value Infectious?

src/infectiousness_manager.rs
```diff
@@ -0,0 +1,263 @@
...
+            return None;
+        }
+    }
+
+    context.get_contact_from_setting(
+        person_id,
+        current_setting,
+        (
+            (Alive, true),
+            (InfectiousStatus, InfectiousStatusValue::Susceptible),
```
confunguido, line 124: Does this mean that if there's at least one susceptible agent, it will be selected? My interpretation is that this is the time to infection attempt, not successful infection. So, we accept/reject the infection attempt, but if not susceptible, then the infection is unsuccessful, and we move to the next attempt. Is that right?

src/rate_fns/mod.rs
```diff
@@ -0,0 +1,157 @@
...
+fn get_fn(context: &Context, index: usize) -> &dyn InfectiousnessRateFn {
+    context.get_data_container(RateFnPlugin).unwrap().rates[index].as_ref()
+}
+
+impl InfectiousnessRateExt for Context {
+    fn add_rate_fn(&mut self, dist: Box<dyn InfectiousnessRateFn>) {
+        let container = self.get_data_container_mut(RateFnPlugin);
+        container.rates.push(dist);
+    }
+    fn assign_random_rate_fn(&mut self, person_id: PersonId) {
```
confunguido, line 82: Is this attempting to incorporate variability in the population? I'm a bit confused about this.

src/infection_propagation_loop.rs
```diff
@@ -0,0 +1,247 @@
...
+/// Eventually, these will actually be loaded from a file
+pub fn load_rate_fns(context: &mut Context) {
+    let parameters = context.get_global_property_value(Parameters).unwrap();
+    let global_transmissibility = parameters.global_transmissibility;
+    let max_time = parameters.infection_duration;
+
+    let create_rate_fn =
+        |rate: f64| Box::new(ConstantRate::new(rate * global_transmissibility, max_time));
+    context.add_rate_fn(create_rate_fn(1.0));
+    context.add_rate_fn(create_rate_fn(2.0));
```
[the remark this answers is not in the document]
k88hudson-cfa: Just to demonstrate that we can load more than 1

src/infection_propagation_loop.rs
```diff
@@ -0,0 +1,247 @@
...
+
+    // Seed the initial population
+    context.add_plan(0.0, move |context| {
+        seed_infections(context, initial_infections);
+    });
+
+    // Add a plan to shut down the simulation after `max_time`, regardless of
+    // what else is happening in the model.
+    context.add_plan(max_time, |context| {
+        context.shutdown();
```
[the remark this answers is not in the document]
k88hudson-cfa: Good question – I thought it was sort of part of the overall loop, but yeah now that you mention it maybe it can go back in main

src/infection_propagation_loop.rs
k88hudson-cfa: I'm just using what was in TransmissionManager before...happy to change it back

src/rate_fns/mod.rs
k88hudson-cfa: Yeah, the purpose of this was to randomly assign one of the available curves to each person. Is that a reasonable way to do it? Or do you want something else here?

src/infection_propagation_loop.rs
confunguido: I lean towards infection status.

src/rate_fns/mod.rs
confunguido: Fine with me

src/infection_propagation_loop.rs
```diff
@@ -0,0 +1,247 @@
...
+    InfectiousStatusValue::Susceptible
+);
+
+define_rng!(InfectionRng);
+
+fn schedule_next_forecasted_infection(
+    context: &mut Context,
+    person: PersonId,
+    get_forecast: impl Fn(&Context, PersonId) -> Option<Forecast> + 'static,
+    evaluate_forecast: impl Fn(&mut Context, PersonId, f64) -> Option<PersonId> + 'static,
```
ekr-cfa, line 27: Why are you passing in function pointers? This can only be invoked internally and thus it always has the same function except in test

src/infection_propagation_loop.rs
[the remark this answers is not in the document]
ekr-cfa: OK, but these are in production code!

src/infectiousness_manager.rs
ekr-cfa: This actually brings up an interesting point:
If you are in a setting with 3 people but 2 of which are  recovered, is the total scaling factor 1^alpha or 3^alpha.

ekr-cfa (changes requested):
Some initial comments.

src/infectiousness_manager.rs
confunguido: I would do this as 3^alpha and sample a random person. Then, check their susceptibility, which can also be reduced by other interventions (e.g., vaccine), and accept/reject accordingly.

src/infection_propagation_loop.rs
```diff
@@ -0,0 +1,247 @@
...
+        trace!("Person {person} has recovered at {current_time}");
+        context.set_person_property(person, InfectiousStatus, InfectiousStatusValue::Recovered);
+        return;
+    }
+    let Forecast {
+        next_time,
+        forecasted_total_infectiousness,
+    } = forecast.unwrap();
+    context.add_plan(next_time, move |context| {
+        let next_contact = evaluate_forecast(context, person, forecasted_total_infectiousness);
```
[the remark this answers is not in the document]
k88hudson-cfa: I think it makes more sense to choose the contact at the end of / after the evaluation step, not during the forecast, but I think we should defer this discussion to when we've worked through and proposed a design for how settings should work with respect to forecasted/current total infectiousness as well as contact assignment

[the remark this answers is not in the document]
k88hudson-cfa: Like for example, you might not be in the setting anymore at the time the forecast is evaluated (like if you start teleworking)

src/infectiousness_manager.rs
```diff
@@ -0,0 +1,260 @@
...
+    }
+
+    // TODO<ryl8@cdc.gov>: Right now, we always get a person that happens to be
+    // alive and susceptible. This should be updated to choose a contact from
+    // a person's setting.
+    context.get_contact(
+        person_id,
+        (
+            (Alive, true),
+            (InfectionStatus, InfectionStatusValue::Susceptible),
```
ChiragKumar9, line 118:
```suggestion
```

k88hudson-cfa: Thanks, makes sense

src/infection_propagation_loop.rs
k88hudson-cfa: I wanted to make this a bit easier to test

src/infectiousness_manager.rs
k88hudson-cfa: Let's address this in the setting design

ekr-cfa (commented):
Overall: please don't do:

```
// Comment
foo
// Comment
bar
```

It's hard to read a block like this.

Instead put in whitespace, like so:

```
// Comment
foo

// Comment
bar
```

src/parameters.rs
```diff
@@ -6,24 +6,19 @@ use serde::{Deserialize, Serialize};
...
+    pub global_transmissibility: f64,
     pub infection_duration: f64,
-    pub generation_interval: f64,
     pub report_period: f64,
     pub synth_population_file: PathBuf,
 }
 
 fn validate_inputs(parameters: &ParametersValues) -> Result<(), IxaError> {
-    if parameters.r_0 < 0.0 {
+    if parameters.global_transmissibility < 0.0 {
```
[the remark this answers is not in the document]
k88hudson-cfa: I'm replace this with `rate_of_infection`, which will be a constant rate for all individuals right now.

src/rate_fns/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+        let max = self.get_data_container_mut(RateFnPlugin).rates.len();
+        let index = self.sample_range(InfectiousnessRateRng, 0..max);
+        self.set_person_property(person_id, RateFnId, Some(index));
+    }
+    fn get_person_rate_fn(&self, person_id: PersonId) -> &dyn InfectiousnessRateFn {
+        let index = self
+            .get_person_property(person_id, RateFnId)
+            .unwrap_or_else(|| panic!("No rate function for {person_id}"));
```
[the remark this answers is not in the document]
k88hudson-cfa: In this model, I personally don't see the advantage of handling this on an individual basis every time I use a rate function – I want to have a a helpful error message and a stack trace.

In practice the place you hit this all the time is in tests when you forgot to set things up

src/infectiousness_manager.rs
```diff
@@ -191,42 +241,38 @@ mod test {
...
-        let p1 = context.add_person((CensusTract, 1)).unwrap();
-        context.set_person_property(p1, TimeOfInfection, Some(OrderedFloat(1.0)));
-        context.add_plan(3.0, move |_| {});
+        let p1 = context.add_person(()).unwrap();
+        context.add_plan(2.0, move |context| {
+            context.infect_person(p1);
+        });
         context.execute();
-        assert_eq!(context.get_elapsed_infection_time(p1), 2.0);
+        let InfectionDataValue::Infected { infection_time, .. } =
```
ekr-cfa, line 251: I wonder if we should just write a macro that knows how to pull one variant out of things without this let/else.

src/infectiousness_manager.rs
```diff
@@ -191,42 +241,38 @@ mod test {
...
+            context.get_person_property(p1, InfectionData)
+        else {
+            panic!("Person {p1} is not infected")
+        };
+        assert_eq!(infection_time, 2.0);
+        context.get_person_rate_fn(p1);
```
ekr-cfa, line 257: Let's add a test for `recover_person`

ekr-cfa (approved):
LGTM minus some small issues

src/infectiousness_manager.rs
k88hudson-cfa: I think this is sa good idea, filed: https://github.com/CDCgov/ixa/issues/220

PR #56, opened 2025-02-13
Add get_params utility

k88hudson-cfa:
This is just a small utility to make getting parameter values a little cleaner and replaces a bunch of `unwrap`s everywhere (which kind of suck without a `message`) 

I also added a test for the default config file that will fail if we forget to update it.

### DIFF
--- a/src/parameters.rs
+++ b/src/parameters.rs
@@ -55,6 +55,16 @@ mod test {
     use crate::parameters::{ContextParametersExt, GlobalParams, Params};
 
     #[test]
+    fn test_default_input_file() {
+        let mut context = Context::new();
+        let path = std::path::Path::new(env!("CARGO_MANIFEST_DIR")).join("input/input.json");
+        context
+            .load_global_properties(&path)
+            .expect("Could not load input file");
+        context.get_params();
+    }
+
+    #[test]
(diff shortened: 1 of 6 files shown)


src/parameters.rs
```diff
@@ -32,44 +32,63 @@ fn validate_inputs(parameters: &ParametersValues) -> Result<(), IxaError> {
...
     use super::validate_inputs;
     use std::path::PathBuf;
 
-    use crate::parameters::ParametersValues;
+    use crate::parameters::{ContextParametersExt, GlobalParams, Params};
 
     #[test]
-    fn test_validate_rate_of_infection() {
-        let parameters = ParametersValues {
+    fn test_default_input_file() {
```
confunguido, line 58: I'm unsure what this is checking. Is it that the get_params() is successful?

confunguido (approved):
This is great and what I wanted :) I was confused about `CARGO_MANIFEST_DIR` but I think it makes sense.

src/parameters.rs
k88hudson-cfa: This is checking that `input/input.json` parses correctly (it will fail if we change parameters and we forget to update it)

PR #58, opened 2025-02-18
Add a way of specifying an empirical rate. Closes #48.

### LINKED ISSUE #48: Implement variable rate function for empirical distributions
`InfectiousnessRateFn` is a trait which is intended to support multiple kinds of rate functions generically; the only one currently implemented is `ConstantRate`.

We'd like to implement some new time-varying rate functions, which can get added alongside `ConstantRate`. Rate functions can get instantiated with any set of parameters and must define some functions which also specify how they are scaled and offset.

Note that the individual rate functions get loaded in `load_rate_fns` (see `infection_propagation_loop.rs`)

We thought  empirical distributions (that could be loaded in from files) might be a good start for this model.

ChiragKumar9:


src/rate_fns/empirical_rate.rs
```diff
@@ -0,0 +1,242 @@
...
+        // We need to use the running cumulative integral to estimate the inverse rate, so calculate
+        // it once and then store it for later.
+        let mut cum_rates = cumulative_integral(&times, &instantaneous_rate)?;
+        // `cum_rates` is one element shorter than `times` because its elements are the integral
+        // from the start of the time series to the time at the same index in `times`. We need to
+        // add the extra infectiousness that is there at the beginning of the timeseries.
+        cum_rates
+            .iter_mut()
+            .for_each(|x| *x += instantaneous_rate[0]);
+        cum_rates.insert(0, instantaneous_rate[0]);
```
ChiragKumar9, line 43: I'm not 100% certain this choice is right. I think the other reasonable choice is to add 0 (the integral from x[0] to x[0] of y is always 0) but this implies that x[0] = 0.0 -- in words, that there is no infectiousness before the first value at which we have a rate. I think this approach accounts for those challenges, but I'm not entirely convinced.

src/rate_fns/empirical_rate.rs
```diff
@@ -0,0 +1,233 @@
...
+        }
+        // We need to use the running cumulative integral to estimate the inverse rate, so calculate
+        // it once and then store it for later.
+        let mut cum_rates = cumulative_integral(&times, &instantaneous_rate)?;
+        // `cum_rates` is one element shorter than `times` because its elements are the integral
+        // from the start of the time series to the time at the same index in `times`. We need to
+        // add the extra infectiousness that is there at the beginning of the timeseries.
+        cum_rates
+            .iter_mut()
+            .for_each(|x| *x += instantaneous_rate[0]);
```
KOVALW, line 42: If we were calculating the relative infectiousness as a function of infectiousness rate, there would be no need for a constant because of the fixed interval, but adding the instantaneous rate at zero here doesn't get around the issue of unknown base constant for total infectiousness because we would have already included this timestamp in the trapezoidal integration method. 

Since we're calculating total infectiousness, we might want to multiply the first observed instantaneous infectiousness rate by the length of time before the observation of instantaneous rates (e.g. if the time infectious prior to a time-varying infectiousness empirical distribution is 1 day or 1 hour, we would have differences in what the base cumulative rate at time 0 being added is). This approach, however, would require the assumption that we have a good idea about the shape of that time period (we assumed that the first observed instantaneous rate is either fixed for some window prior to t=0 or has some increase from 0 to `instantaneous_rate[0]`) and would require at least one additional parameter that is very challenging to have good data to inform. I would therefore suggest that we take the relative infectiousness approach here and not add a constant (which assumes that our first observed instantaneous rate is the rate at the time when infectiousness begins).

KOVALW (changes requested):
Just a couple of thoughts on the sample points for the empirical distribution - there's an implicit assumption that the sample points are evenly spaced and that the first observed instantaneous infectiousness rate is the same at the time of first rate observation $t$ and at $t-1$. The first should either be made explicit through function name changes and a check for even spacing in `new` or should be removed by adding corrective factors during integration. The second should be simply removed, and in my opinion, replaced by the assumption that infectiousness prior to observation of the first instantaneous rate is 0. 

I do think there's plenty of room for discussion here! The methods otherwise do look suitable

src/rate_fns/empirical_rate.rs
```diff
@@ -0,0 +1,233 @@
...
+    }
+    fn cum_rate(&self, t: f64) -> f64 {
+        // Integrate rate function up until lower index -- over all times in the samples of the rate
+        // function less than t.
+        let lower_index = get_lower_index(&self.times, t);
+        let mut cum_rate = self.cum_rates[lower_index];
+        // Now we need to estimate the extra area from the last time in our samples of the rate
+        // function to t
+        // Estimate the rate at t
+        let estimated_rate = linear_interpolation(
```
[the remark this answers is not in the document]
ChiragKumar9: For the reason you describe, this would result in an extra calculation of `lower_index`, which is not only redundant but bad since that is by far the more computationally-intensive part of the procedure. I think there's a way to refactor the code to write a function that outputs both lower index and the interpolated value, but that feels more confusing than this current structure.

src/rate_fns/empirical_rate.rs
KOVALW: Put some thoughts on this in my review. In short I think the second solution is the right path here

src/utils/midpoint_integration.rs
```diff
@@ -0,0 +1,103 @@
...
+/// - If `x` and `y` do not have the same length
+pub fn midpoint_integration(x: &[f64], y: &[f64]) -> Result<f64, IxaError> {
+    if x.len() != y.len() {
+        return Err(IxaError::IxaError(
+            "`x` and `y` must have the same length.".to_string(),
+        ));
+    }
+    Ok(x.windows(2)
+        .zip(y.windows(2))
+        .map(|(x, y)| (x[1] - x[0]) * (y[0] + y[1]) / 2.0)
```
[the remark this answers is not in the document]
ChiragKumar9: I don't follow at all. Can you please explain? Nothing in this rule requires that the x values are evenly sampled. This is just implementing the trapezoidal integration rule for finite data samples, and that rule happens to be the midpoint integration rule when you also have finite sample data and need to do interpolation between points.

src/rate_fns/empirical_rate.rs
ChiragKumar9: I don't follow fully. But I believe the method you are describing does not hold up for the case where the provided instantaneous rate at `t = 0.0` is not 0.

KOVALW: Even if the instantaneous rate at $t = 0.0$ is not zero, we would still want the integral to be zero when evaluated at that time. Under the trapezoidal rule to integrate some rate function $r(t)$ and obtain $R(t)$, the first time point we evaluate is $t = \Delta t$, where $R(\Delta t)$ is approximately $0.5 \Delta t [r(0) + r(\Delta t)]$ and $R(0)=0$, regardless of $r(0)$.

ChiragKumar9: I think we should have a call about this! This is exactly where my head was at but then I thought through the case where the first time value is not 0.0 and that convinced me we need to do it a different way. When the instantaneous rate at t = 0 is 0, what's implemented now reduces to what you describe.

ChiragKumar9: I refactored so that having the integral be 0 at time = 0 is the only logical choice. Thanks for your thoughts here!

src/rate_fns/empirical_rate.rs
```diff
@@ -0,0 +1,369 @@
...
+    /// Helper function to return all the elements we may need from the rate function interpolation
+    /// routine.
+    fn lower_index_and_rate(&self, t: f64) -> (usize, usize, f64) {
+        // Find indeces of times that window `t`
+        let (integration_index, interpolation_index) = get_lower_index(&self.times, t);
+        // Linear interpolation between those two points
+        (
+            integration_index,
+            interpolation_index,
+            linear_interpolation(
```
KOVALW, line 97: **Subjective** suggestion: since the integration index is either always equal to the interpolation index or one greater than the interpolation index (extrapolation), it might make more sense to switch the order of these. It might be more readable has well to have more frequent `(_, integration_index, rate)` than `(index, _, rate)` assignments.

KOVALW (approved):
LGTM! The thorough testing was really helpful to see expected behavior and how the pieces fit together. Nice work on refactoring

src/rate_fns/empirical_rate.rs
```diff
@@ -0,0 +1,371 @@
...
+    }
+    #[allow(dead_code)]
+    /// Helper function to get out the cumulative rates at each time.
+    fn get_cum_rates(&self) -> Vec<f64> {
+        self.cum_rates.clone()
+    }
+    /// Helper function to return all the elements we may need from the rate function interpolation
+    /// routine.
+    fn lower_index_and_rate(&self, t: f64) -> (usize, usize, f64) {
+        // Find indeces of times that window `t`
```
RobertJacobsonCDC, line 91: I've noticed we use "indexes" pretty much everywhere instead of "indices." (But "indeces" is a typo.)

src/rate_fns/empirical_rate.rs
```diff
@@ -0,0 +1,371 @@
...
+            cum_rates,
+            ..empirical_rate_no_cum
+        })
+    }
+    #[allow(dead_code)]
+    /// Helper function to get out the cumulative rates at each time.
+    fn get_cum_rates(&self) -> Vec<f64> {
+        self.cum_rates.clone()
+    }
+    /// Helper function to return all the elements we may need from the rate function interpolation
```
RobertJacobsonCDC, line 88: Yeah, but tell me what these elements are. I don't know what the returned numbers `(usize, usize, f64)` represent by looking at the function signature, maybe not even by looking at the body of the function, so you should at least tell me in the doc comment.

Is it ever the case that the `xs` values (`EmpiricalRate::times` values) are *not* distinct? What I'm thinking is, if we have to deal with duplicated values, then what we are really wanting from this function and it's helper `get_lower_index()` is the two _distinct_ points that are most appropriate to use for linear interpolation for `xp` (`t` in this function). So if, for example, we have a weird case like `xs = [..., 2.0, 3.0, 3.0]`, then you really want `x1 = 2.0` and `x2 = 3.0` (and whatever the corresponding `y` values are). Right?

If the times / `xs` are always distinct, it doesn't really matter, and we should just ignore this theoretical edge case anyway, IMO. But you mentioned this case in the doc comment of `get_lower_index()`, so....

RobertJacobsonCDC (changes requested):

src/rate_fns/empirical_rate.rs
ChiragKumar9: Thanks. I've revised the function documentation throughout for clarity.

With regards to your question, there is never a case of an empirical rate where the times values are not distinct biologically speaking. But, it's worth noting that we don't explicitly check for this when making an empirical rate struct (regretfully, checking for distinct values is not trivial because floats don't implement `Hash` or `Ord` ah!), so someone could erroneously make that sort of empirical rate struct. That being said, I think the way the integration and interpolation functions are written that they can handle this case in a reasonable way.

What I do think is a problem is a slightly more in depth case which is what I reference in the doc comments of `linear_interpolation`: consider that the instantaneous rate is 0 over some period. This means that the cumulative rate will be a constant value over that period. We need to invert the cumulative rate in `inverse_cum_rate`, so we may end up with a situation where the two x values are the same (same value of cumulative rate) but the y values (times) are different. I think what you're describing as an alternative is reasonable, but I'm not sure it's worth the extra run through of a binary search algorithm for this corner case. I tend to believe that we should do the interpolation over the smallest window possible (neighboring cumulative rates values) even if that requires an approximation for the corner case simply because we are more likely to be accurate (even with an approximation) over small windows.

src/rate_fns/empirical_rate.rs
```diff
@@ -0,0 +1,402 @@
...
+        )?;
+        // `cum_rates` is one element shorter than `times` because its elements are the integral
+        // from the start of the time series to the time at the same index in `times` (omitting the
+        // integral from the first value in the time series to itself).
+        // We add that zero here:
+        cum_rates.insert(0, 0.0);
+        // Next we account for there being infectiousness potentially before the first value in the
+        // timeseries -- i.e., if the rate timeseries does not start at 0.
+        // We need to calculate that infectiousness and add it to our cumulative rates.
+        let (_, _, estimated_rate_zero) = empirical_rate_no_cum.lower_index_and_rate(0.0);
```
RobertJacobsonCDC, line 70: The estimated rate at `t=0` can be negative in this implementation even when all other instantaneous rates are nonnegative—in fact I think this is always the case for any function that is strictly convex in a neighborhood of `t=0` containing the first sampled time, because the interpolation is linear. Consider:

```rust
// y = x - 0.5;
let xs = vec![1.0, 2.0, 3.0];
let ys = vec![0.5, 1.5, 2.5];
let empirical_rate = EmpiricalRate::new(xs, ys).expect("failed to make");
```

Compare with what you get with:

```rust
let xs = vec![0.0, 1.0, 2.0, 3.0];
let ys = vec![0.0, 0.5, 1.5, 2.5];
let empirical_rate = EmpiricalRate::new(xs, ys).expect("failed to make");
```

As I understand it, a negative instantaneous hazard rate is not allowed in this model, so I think you end up underestimating the cumulative rates. Right? Or is this ok?

src/rate_fns/empirical_rate.rs
RobertJacobsonCDC: > so we may end up with a situation where the two x values are the same (same value of cumulative rate) but the y values (times) are different.

In this case, aren't the semantics of `inverse_cum_rate()` that it returns the smallest time? That's how I interpret the doc comment for `inverse_cum_rate()` anyway. Whether that's the "right" thing to do is a different questions, I guess.

RobertJacobsonCDC (changes requested):

src/rate_fns/empirical_rate.rs
ChiragKumar9: Can you help me understand why it records the smallest time? Because we're doing binary search in getting the indices, a deterministically random index should be returned.

src/rate_fns/empirical_rate.rs
ChiragKumar9: Thanks! This is such a good point that I did not catch. I thought I had a test for this exact idea but I realized that I had done what was correct in the inverse cumulative rate calculation but not the cumulative rate calculation. Added a test for this as well.

src/rate_fns/empirical_rate.rs
RobertJacobsonCDC: I'm not describing how you are implementing `inverse_cum_rate()` but rather how it is described in the docs on the trait:

> Returns the expected time, starting at 0, at which a number of infection `events` will have occurred.
>
> E.g., Where t=day, `inverse_cum_rate(6.0)` -> 2.0 means that we would expect that it would take 2 days to infect 6 people.

It would be odd if this meant, "We would expect that it would take 2 days to infect 6 people AND ALSO 3.5 days to infect 6 people." The phrase "starting at 0" suggests to me that `inverse_cum_rate()` is intended to give the first time at which `n` infections (are expected to) occur. Maybe this meaning wasn't intended.

In any case, if you think the code is doing what it should be doing in this PR, who am I to argue?

RobertJacobsonCDC (approved):
You haven't clicked "resolve" on the conversations—which is fine—so as long as you have gone through them all and are satisfied that this PR makes sense for your model, then it looks good to me, too.

src/rate_fns/empirical_rate.rs
ChiragKumar9: Thanks for raising this point and examining the code this carefully! I thought about this some more and you are absolutely right -- there is a bit of a discordance here. I'm going to implement a "left search" check on the binary search here to make sure we always return the smallest time at which this is true. Appreciate the catch here!

src/rate_fns/empirical_rate.rs
ChiragKumar9: For posterity, since the interpolation index depends on the integration index, I think the integration index should come first! Since one can always calculate the interpolation index from the integration index but not vice versa.

src/rate_fns/empirical_rate.rs
```diff
@@ -146,14 +149,35 @@ impl InfectiousnessRateFn for EmpiricalRate {
...
+        // extra_events = rate(time_passed) * t + t^2 * slope / 2 = (rate(time_passed) + t * slope / 2) * t
+        // 0 = t^2 * slope / 2 + rate(time_passed) * t - extra_events
+        let delta_y = self.instantaneous_rate[interpolation_index + 1]
+            - self.instantaneous_rate[interpolation_index];
+        let delta_x = self.times[interpolation_index + 1] - self.times[interpolation_index];
+        let slope = delta_y / delta_x;
+        // Because extra_events is >= 0 by definition, we know that this quadratic equation will
+        // always have at least one root. So, we don't need to worry about the case where this
+        // function returns `None`.
+        let upper_root = upper_quadratic_root(
```
RobertJacobsonCDC, line 209: Let $x$ = `extra_events` > 0 , $r_{t_p}$ = `rate(time_passed)` > 0, and $m =$ slope. Your calculation shows

```math
\frac{m}{2}t^2 + r_{t_p} t - x = 0.
```

Recall the quadratic formula, which gives the roots of $a t^2 + bt + c = 0$:

```math
t = \frac{-b \pm \sqrt{b^2-4ac}}{2a}
```

If real roots exist, $b^2-4ac \geq 0$, and so $\sqrt{b^2-4ac} \geq 0$. Substituting coefficients:

```math
t = \frac{-r_{t_p} \pm \sqrt{r_{t_p}^2 + 2mx}}{m}.
```

If $m>0$, $\displaystyle t = \frac{\sqrt{r_{t_p}^2 + 2mx} -r_{t_p}}{m}.$

If $m < 0$, $\displaystyle t = \frac{-\left(r_{t_p}+\sqrt{r_{t_p}^2 + 2mx}\right)}{m}.$

RobertJacobsonCDC: So you shouldn't need `upper_quadratic_root()` or the `roots` dependency.

RobertJacobsonCDC (changes requested):

src/rate_fns/empirical_rate.rs
ChiragKumar9: You are right in the case where `m > 0`. But, for a constant rate, `m = 0`, so we don't have a quadratic anymore. This corner case was accounted for in `roots`. That being said, I think your suggestion makes the code more explicit, so I will still implement the true math! Thanks for pointing this out!

RobertJacobsonCDC: If $m = 0$, then $t = x / r_{t_p}$. The case that both $m = 0$ and $r_{t_p}=0$ is handled earlier in the function, I think.

src/rate_fns/empirical_rate.rs
```diff
@@ -0,0 +1,494 @@
...
+        // We know that events are the integral of rate over time, so we need to solve the following
+        // for t:
+        // extra_events = integral_{time_passed}^{time_passed+t}, rate(\tau) d\tau)
+        // We assume that rates change linearly over adjacent time windows.
+        // extra_events = integral_{0}^{t}, (rate(time_passed) + slope*\tau) d\tau)
+        // extra_events = rate(time_passed) * t + t^2 * slope / 2 = (rate(time_passed) + t * slope / 2) * t
+        // 0 = t^2 * slope / 2 + rate(time_passed) * t - extra_events
+        let delta_y = self.instantaneous_rate[integration_index + 1]
+            - self.instantaneous_rate[integration_index];
+        let delta_x = self.times[integration_index + 1] - self.times[integration_index];
```
confunguido, line 200: Shouldn't this be delta_t?

src/rate_fns/empirical_rate.rs
confunguido: if m is 0, doesn't that just simplify the equation to r_{t_p} -x = 0?

confunguido (commented):
Looks great! A few comments.

src/rate_fns/empirical_rate.rs
ChiragKumar9: Yep! I added that case to return early. And then it turns out you can simplify even once more because the quadratic equation solution arises since we are inverting a quadratic function, and that inverse can really only ever have one solution (i.e., one arm of the quadratic solution). Writing out the math here helped a lot to simplify code structure. Thanks for looking at this in depth!

src/rate_fns/empirical_rate.rs
ChiragKumar9: I also added a test for the two case -- slope is positive and slope is negative -- to make sure we are using the right solution in both cases.

src/rate_fns/empirical_rate.rs
```diff
@@ -203,15 +200,24 @@ impl InfectiousnessRateFn for EmpiricalRate {
...
+                    + extra_events / self.instantaneous_rate[integration_index],
+            );
+        }
+        let discriminant =
+            f64::powi(self.instantaneous_rate[integration_index], 2) + 2.0 * slope * extra_events;
+        // Finally, the reason this quadratic solution arises is because we are inverting a
+        // quadratic function. Therefore, we can show that we always want the solution where we add
+        // the square root of the discriminant because when we make the quadratic function, we are
+        // *accumulating* extra area.
+        let t = (-self.instantaneous_rate[integration_index] + f64::sqrt(discriminant)) / slope;
```
RobertJacobsonCDC, line 219: Let $x =$ `extra_events `, $r_{t_p} =$ `self.instantaneous_rate[integration_index]` and $d = r_{t_p}^2 - 2mx$ (the discriminant). 

When $m>0$, then the largest root is $\displaystyle t = \frac{-r_{t_p} + \sqrt{d}}{m}$, which is what you have here. But when $m < 0$, the largest root is $\displaystyle t = \frac{-r_{t_p}-\sqrt{d}}{m}.$ This is because (starting with this root)

```math
\begin{align*}
\frac{-r_{t_p} - \sqrt{d}}{m} &= \frac{\left|-\left(r_{t_p} + \sqrt{d}\right)\right|}{|m|} \\
&= \frac{\left|\sqrt{d} + r_{t_p}\right|}{|m|} \\
&\geq \frac{\left|\left|\sqrt{d}\right| - \left|r_{t_p}\right|\right|}{|m|} \qquad \text{("reverse" triangle ineq.)} \\
&= \left|\frac{\left|\sqrt{d}\right| - \left|r_{t_p}\right|}{m}\right| \\
&\geq \frac{\left|\sqrt{d}\right| - \left|r_{t_p}\right|}{m} \\
&= \frac{\sqrt{d} - r_{t_p}}{m}. \qquad \text{(both already positive)}
\end{align*}
```

which is the other root.

RobertJacobsonCDC (changes requested):

src/rate_fns/empirical_rate.rs
ChiragKumar9: Thanks, Robert. I was previously wrong about us always needing the upper root. I added a test for the case were $m < 0$ (see `test_inverse_cum_rate_plateaus`), and in both cases you want the root where you add the discriminant. In the case where $m < 0$, the other root is too big. I don't have a deeply good way of thinking about this intuitively, but as I see it, the quadratic equation solution arises in the first place because we are inverting a quadratic function. But, we're really only inverting half of a $x^2$ like thing (the half where $x>0$, so we want our inverse to be $\sqrt{x}$, not $-\sqrt{x}$. I'm not sure if this is a helpful explanation. Happy to also discuss live.

ChiragKumar9: Also `test_inverse_cum_rate_in_bounds`.

RobertJacobsonCDC: So my math is right, but Chirag pointed out that we actually don't want to always take the largest root. In the case that `m < 0`, we want the smallest root. Picture proof attached for posterity. 
IMG_8988

RobertJacobsonCDC (approved):

confunguido (approved):
LGTM

PR #59, opened 2025-02-20
Change `InfectionStatusValue::Infected` to be `InfectionStatusValue::Infectious`

ChiragKumar9:
"Infectious" implies that the person is capable of transmitting disease while "infected" simply says that the person has succumbed to infection. We are interested in when people transmit disease (which may be disconnected from when they "recover" in terms of symptoms), so we want to use "infectious" to better indicate that this status refers to people who are actively capable of infecting others, regardless of their other clinical statuses.

### DIFF
--- a/src/infection_propagation_loop.rs
+++ b/src/infection_propagation_loop.rs
@@ -55,9 +55,9 @@ pub fn load_rate_fns(context: &mut Context) {
     )));
 }
 
-/// Seeds the initial population with a number of infected people.
+/// Seeds the initial population with a number of infectious people.
 fn seed_infections(context: &mut Context, initial_infections: usize) {
-    // First, seed an infected population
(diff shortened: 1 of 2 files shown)


eqmooring (approved):
I think this is reasonable; your comment makes sense. To the extent that the state in question is, in practical modelling terms, the state from which it is necessary to be entertaining the possibility of onward transmission (that's probably bad phrasing, but hopefully roughly in line with what's happening), then I agree that "infectiousness" is the more accurate term rather than "infected". I think were things get tricky is if you have, say, a piecewise intrinsic infectiousness (generation interval) function that is shaped such that for a portion of it the potential for onward transmission is zero, then describing people in this state as "actively capable" of infecting others is at best imprecise.

confunguido (approved):
Looks good to me

PR #60, opened 2025-02-20
Add statistical tests for infection loop. Closes #52.

### LINKED ISSUE #52: Add statistical tests for the model's infection propagation loop
It would be nice to have a test that actually verifies that the rate of infection is what we expect.

What I was thinking here was given a set of hard-coded parameters, is the number of infections we generate within some range of expected values

ChiragKumar9:
This is inspired by the tests we had the previous transmission manager (see #51's deleted files) but welcome suggestions for other statistical tests to include.

### DIFF
--- a/src/infection_propagation_loop.rs
+++ b/src/infection_propagation_loop.rs
@@ -101,12 +101,12 @@ mod test {
 
     use super::seed_infections;
 
-    fn setup_context() -> Context {
+    fn setup_context(seed: u64) -> Context {
         let mut context = Context::new();
         let parameters = Params {
             initial_infections: 3,
             max_time: 100.0,
-            seed: 0,
+            seed,
             rate_of_infection: 1.0,
             infection_duration: 5.0,
             report_period: 1.0,
@@ -122,7 +122,7 @@ mod test {
 
     #[test]
     fn test_seed_infections() {
-        let mut context = setup_context();
+        let mut context = setup_context(0);
         for _ in 0..10 {
             context.add_person(()).unwrap();
         }
@@ -136,7 +136,7 @@ mod test {
 
     #[test]
     fn test_init_loop() {
-        let mut context = setup_context();
+        let mut context = setup_context(0);
         for _ in 0..10 {
             context.add_person(()).unwrap();
         }
--- a/src/infection_propagation_loop.rs
+++ b/src/infection_propagation_loop.rs
@@ -88,10 +88,11 @@ pub fn init(context: &mut Context) {
 #[cfg(test)]
 #[allow(clippy::float_cmp)]
 mod test {
-    use std::path::PathBuf;
+    use std::{cell::RefCell, path::PathBuf, rc::Rc};
 
     use ixa::{
         Context, ContextGlobalPropertiesExt, ContextPeopleExt, ContextRandomExt, ExecutionPhase,
+        PersonPropertyChangeEvent,
     };
     use statrs::assert_almost_eq;
 
@@ -175,6 +176,40 @@ mod test {
                 .query_people((InfectionStatus, InfectionStatusValue::Recovered))
                 .is_empty(),
             "Expected some people to recover"
+        );
+    }
+
+    #[test]
+    fn test_zero_rate_no_infections() {
+        let mut context = setup_context(0, 0.0);
+        for _ in 0..=context
+            .get_global_property_value(Parameters)
+            .unwrap()
+            .initial_infections
+        {
+            context.add_person(()).unwrap();
+        }
+
+        init(&mut context);
+
+        let num_new_infections = Rc::new(RefCell::new(0usize));
+        let num_new_infections_clone = Rc::clone(&num_new_infections);
+        context.subscribe_to_event::<PersonPropertyChangeEvent<InfectionStatus>>(
+            move |_context, event| {
+                if event.current == InfectionStatusValue::Infected {
+                    *num_new_infections_clone.borrow_mut() += 1;
+                }
+            },
+        );
+
+        context.execute();
+
+        assert_eq!(
+            *num_new_infections.borrow(),
+            context
+                .get_global_property_value(Parameters)
+                .unwrap()
+                .initial_infections
         );
     }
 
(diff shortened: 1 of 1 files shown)


src/infection_propagation_loop.rs
```diff
@@ -171,4 +180,131 @@ mod test {
...
+        // to stop those simulations at the end of 1.0 time units and compare the number of infected
+        // people to the infectious rate we used to set up the simulation.
+        // We're also going to check the times at which they are infected -- since we only record
+        // infection times of when people are actually infected, we expect the times to be uniform
+        // on [0, 1].
+        const NUM_SIMS: usize = 10000;
+        let rate = 1.5;
+        // We need the total infectiousness multiplier for the person.
+        let mut total_infectiousness_multiplier = None;
+        let mut num_infections_end_one_time_unit = Vec::<usize>::with_capacity(NUM_SIMS);
```
RobertJacobsonCDC, line 226: The elements of this vector just get summed up at the end, so why not just call this a `usize`? Then you'd have

```rust
// line 226
let mut num_infections_end_one_time_unit = 0;
// line 275
num_infections_end_one_time_unit += infected_count;
// line 279-280
let avg_number_infections =
            num_infections_end_one_time_unit as f64 / NUM_SIMS as f64;
```

RobertJacobsonCDC (approved):
I'd make `num_infections_end_one_time_unit` a `usize`, but otherwise it looks fine to me.

src/infection_propagation_loop.rs
ChiragKumar9: Oh nice catch thanks!

src/infection_propagation_loop.rs
```diff
@@ -171,4 +180,130 @@ mod test {
...
+        // Where we store the infection times.
+        let infection_times = Rc::new(RefCell::new(Vec::<f64>::new()));
+        for seed in 0..NUM_SIMS {
+            let infection_times_clone = Rc::clone(&infection_times);
+            let mut context = setup_context(seed.try_into().unwrap(), rate);
+            // We only run the simulation for 1.0 time units.
+            context.add_plan_with_phase(1.0, ixa::Context::shutdown, ExecutionPhase::Last);
+            // Add a bunch of people so we mimic an infinitely large population.
+            for _ in 0..100 {
+                context.add_person(()).unwrap();
```
confunguido, line 236: Why do you need to unwrap() here?

src/infection_propagation_loop.rs
```diff
@@ -171,4 +180,130 @@ mod test {
...
+
+    #[test]
+    fn test_number_timing_infections_one_time_unit() {
+        // Does one infectious person generate the number of infections as expected by the rate?
+        // We're going to run many simulations that each start with one person infected and have a
+        // very large number of susceptible people (mirroring one susceptible in an infinitely-large
+        // population) but those people themselves cannot transmit secondary infections. We're going
+        // to stop those simulations at the end of 1.0 time units and compare the number of infected
+        // people to the infectious rate we used to set up the simulation.
+        // We're also going to check the times at which they are infected -- since we only record
```
confunguido, line 219: Can you elaborate here why infection times are expected to be uniform?

src/infection_propagation_loop.rs
```diff
@@ -171,4 +180,130 @@ mod test {
...
+            let infection_times_clone = Rc::clone(&infection_times);
+            let mut context = setup_context(seed.try_into().unwrap(), rate);
+            // We only run the simulation for 1.0 time units.
+            context.add_plan_with_phase(1.0, ixa::Context::shutdown, ExecutionPhase::Last);
+            // Add a bunch of people so we mimic an infinitely large population.
+            for _ in 0..100 {
+                context.add_person(()).unwrap();
+            }
+            // We don't want infectious people beyond our index case to be able to transmit, so we
+            // have to do setup on our own since just calling `init` will trigger a watcher for
```
confunguido, line 239: Tbh, I don't like this but can't think of anything better.  Is the goal to keep track of the number of infection attempts as in the expected number of infections in a homogeneous and large completely susceptible population? What's the difference with having two people, one infected, the other one recovered. Then infectious person would keep trying to infect the recovered person over and over again, but they wouldn't be able to transmit to others. Not suggesting to implement this but wondering if there's a more explicit/straightforward way of measuring infection attempts.

src/infection_propagation_loop.rs
```diff
@@ -171,4 +180,130 @@ mod test {
...
+        context.execute();
+
+        assert_eq!(
+            *num_new_infections.borrow(),
+            context.get_params().initial_infections
+        );
+    }
+
+    #[test]
+    fn test_number_timing_infections_one_time_unit() {
```
confunguido, line 212: I like the idea of this test and would also like to see it as a plot outside of ixa, which I think we can recreate with the transmission chain report.

src/infection_propagation_loop.rs
ChiragKumar9: `context.add_person(_)` returns `Result<PersonId, IxaError>` to return an error in case there are person properties that need to be initialized at start-up that aren't specified!

src/infection_propagation_loop.rs
ChiragKumar9: Hmmm, I think this is a great point. You are absolutely right about the goal. I really like your suggestion instead, and that's what I implemented  before. I am happy to switch to that implementation (and I think it will be faster because the population size is smaller) but do you think the code is cleaner?

https://github.com/CDCgov/ixa-epi-isolation/blob/9ee4b69386f8a4cdee5299f4e345d85b04d66dc5/src/transmission_manager.rs#L355

ChiragKumar9: This ended up being so much faster, thanks Guido!

src/infection_propagation_loop.rs
ChiragKumar9: Addressed in a revised version of the comment. Let me know if the new text clarifies.

confunguido (approved):
lgtm

PR #62, opened 2025-02-27
Enable reading in a library of rate functions from an external file.

ChiragKumar9:
Read in a single rate function as a global parameter, or specify a file to read in a library of empirical rate functions. Have more thoughts for how we can further generalize this to allow reading in a library of any type of rate functions, but I think that should go in a separate PR.


src/parameters.rs
```diff
@@ -4,6 +4,12 @@ use std::path::PathBuf;
 use ixa::{define_global_property, ContextGlobalPropertiesExt, IxaError};
 use serde::{Deserialize, Serialize};
 
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
```
ChiragKumar9, line 7: @k88hudson-cfa what do you think of this approach for specifying in different rate functions?

src/infection_propagation_loop.rs
```diff
@@ -39,20 +41,22 @@ fn schedule_recovery(context: &mut Context, person: PersonId) {
...
-    } = context.get_params();
-
-    context.add_rate_fn(Box::new(ConstantRate::new(
-        rate_of_infection,
-        infection_duration,
-    )));
+/// Instantiate the rate functions specified in the global parameter `rate_of_infection` as actual
+/// rate functions for the simulation that are assigned randomly to agents when they are infected.
+pub fn instantiate_rate_fns(context: &mut Context) -> Result<(), IxaError> {
+    let rate_of_infection = context.get_params().rate_of_infection.clone();
```
ChiragKumar9, line 47: Really don't love that this structure basically duplicates the info in `rate_of_infection`. I think there are two potential solutions: (a) implement `InfectiousnessRateFn` on `enum Rates` instead, or (b) actually move the info in `rate_of_infection` rather than copying it. I don't think (a) is a possible option because in implementing the methods for `InfectiousnessRateFn` on `EmpiricalRate`, you need to create new vectors that you store in the initialization process, and I don't see a way to do this if we just implement the method on the type as it is fed in from deserialization. I think (b) is an easy option, but I am open to other thoughts and ways of thinking about this.


I think we have to move the info (at least) because in `EmpiricalRate` we want to create

src/infection_propagation_loop.rs
```diff
@@ -39,20 +41,22 @@ fn schedule_recovery(context: &mut Context, person: PersonId) {
...
+    let infection_duration = context.get_params().infection_duration;
+
+    for rate in rate_of_infection.clone() {
+        context.add_rate_fn(match rate {
+            Rates::Constant(rate) => Box::new(ConstantRate::new(rate, infection_duration)?),
+            Rates::Empirical(rate_fn) => {
+                let (t, r): (Vec<f64>, Vec<f64>) = rate_fn.into_iter().unzip();
+                Box::new(EmpiricalRate::new(t, r)?)
```
ChiragKumar9, line 55: Also, doesn't using `Box` create an overhead here/should we be concerned about that? I could imagine us having a number of infectiousness curves on the order of the number of agents in the simulation.

k88hudson (approved):
Nice, this is much better. LGTM 👍

k88hudson-cfa (approved):

confunguido (approved):
Only comment is that it's not possible to know what the `infectiousness_rate_fn` parameters are for the constant case in the input parameters. It would be nice to define it as `{rate: 1.0, infectiousness_period: 5}`. Mainly because it's not obvious that infectiousness period has to be a parameter for the rate function.

PR #63, opened 2025-02-28
Integrate infection duration into `InfectiousnessRateFn`. Proposal to close #55.

### LINKED ISSUE #55: Implement more sophisicated recovery/symptom process?
Right now in the model, a person's recovery is scheduled according to the global infection duration parameter.

Do we want to handle this in some more sophisticated way?

ChiragKumar9:
- Although infection duration is still specified as a global parameter, this structure enables each rate function type to have their own definition of infection duration and for each rate function instance to have a different value of the infection duration.
- If we switch to the structure in #62, infection duration can be specified for each rate function at input.
- This PR does not solve the problem of needing to model _symptom_ recovery, and a subsequent PR will introduce a `ClinicalDiseaseSymptomStatus` and tracker.

### DIFF
--- a/src/rate_fns/empirical_rate.rs
+++ b/src/rate_fns/empirical_rate.rs
@@ -509,4 +509,15 @@ mod test {
         // greater than 3.0 because the rate starts moving again after t = 3.0.
         assert!(empirical.inverse_cum_rate(1.6).unwrap() > 3.0);
     }
+
+    #[test]
+    fn test_infection_duration_remaining() {
+        let empirical = EmpiricalRate::new(
+            vec![0.0, 1.0, 2.0, 3.0, 4.0, 5.0],
+            vec![1.0, 1.0, 0.0, 0.0, 1.0, 0.0],
+        )
+        .unwrap();
+        assert_almost_eq!(empirical.infection_duration_remaining(2.0), 3.0, 0.0);
+        assert_almost_eq!(empirical.infection_duration_remaining(7.0), -2.0, 0.0);
+    }
 }
--- a/src/rate_fns/empirical_rate.rs
+++ b/src/rate_fns/empirical_rate.rs
@@ -144,7 +144,7 @@ impl InfectiousnessRateFn for EmpiricalRate {
     }
 
     fn inverse_cum_rate(&self, events: f64) -> Option<f64> {
-        if events > *self.cum_rates.last().unwrap() {
+        if events > self.cum_rates[self.cum_rates.len() - 1] {
             return None;
         }
         // We want to return the time at which `events` would have happened. At a high level, this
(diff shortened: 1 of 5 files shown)


src/rate_fns/empirical_rate.rs
```diff
@@ -219,6 +219,10 @@ impl InfectiousnessRateFn for EmpiricalRate {
         let t = (-self.instantaneous_rate[integration_index] + f64::sqrt(discriminant)) / slope;
         Some(self.times[integration_index] + t)
     }
+
+    fn infection_duration_remaining(&self, t: f64) -> f64 {
```
confunguido, line 223: Maybe I'm missing something but why infection_duration_remaining as opposed to infection_duration?

src/rate_fns/empirical_rate.rs
```diff
@@ -505,4 +509,15 @@ mod test {
         // greater than 3.0 because the rate starts moving again after t = 3.0.
         assert!(empirical.inverse_cum_rate(1.6).unwrap() > 3.0);
     }
+
+    #[test]
+    fn test_infection_duration_remaining() {
+        let empirical = EmpiricalRate::new(
+            vec![0.0, 1.0, 2.0, 3.0, 4.0, 5.0],
+            vec![1.0, 1.0, 0.0, 0.0, 1.0, 0.0],
```
confunguido, line 517: What happens if the last index isn't 0.0?

src/rate_fns/empirical_rate.rs
ChiragKumar9: `_remaining` because the method is a function of `t` (time since infection began), and the method does not just spit out the total infection duration but rather the infection duration _remaining_ (total infection duration would be `t + self.infection_duration_remaining(t)`. This structure is more general and robust to potentially needing to expand the infection duration during an infection.

src/rate_fns/empirical_rate.rs
ChiragKumar9: Thanks. The `infection_duration_remaining` method is agnostic to the value of `instantaneous_rate`, so it makes no difference: https://github.com/CDCgov/ixa-epi-isolation/blob/06294dd8842d3d51d64c9a3071a1bb1721da2660/src/rate_fns/empirical_rate.rs#L224.

More generally, I think you are asking the question "can someone be infectious at the last time point for which we provide a rate value", and the answer is yes: https://github.com/CDCgov/ixa-epi-isolation/blob/06294dd8842d3d51d64c9a3071a1bb1721da2660/src/rate_fns/empirical_rate.rs#L488 and many more examples.

In the `EmpiricalRate` struct, I'm defining the convention that the last provided time point is the infection duration, but there's no reason that someone couldn't write their own wrapper on the struct where the infection duration was the last time plus some random variable that's a function of the rate at the last time point (as is the case for a constant hazard rate function).

I'm happy to change the test if you'd like, but I actually wanted to purposefully put in a 0.0 as the last rate value to show someone who was reading the tests that the infection duration is defined as the last time point, not the time at which the rate hits 0 for the last time. This means that if someone wanted to draw out the infection duration (i.e., time spent in `Infectious` compartment) in a trivial way, they could just append a time and a rate of 0.0 to their `EmpiricalRate` vectors. What do you think?

confunguido (approved):

PR #70, opened 2025-03-11
Sketching setting creation with some properties closes #47

### LINKED ISSUE #47: Implement settings
This needs some design work / addition details, but the overall concept is that we want to support specific ratios and/or periods of time (e.g., you are at work 9-5) for specific settings, and allow people to have different itineraries (determined by their properties)

confunguido:
This PR contains the definition of settings with the following functionality:

- register setting types (e.g., Home, Work)
- adding itineraries for a person
- calculating total infectiousness multiplier for a person based on itinerary
- get current members of a setting
- get a contact from a setting

Todo:
- incorporate into model
- expand or add another way to get a contact to first choose the setting to draw the contact from
- add more tests


src/settings.rs
```diff
@@ -0,0 +1,607 @@
...
+        container
+            .setting_types
+            .insert(TypeId::of::<T>(), Box::new(setting_type));
+
+        // Add properties
+        container
+            .setting_properties
+            .insert(TypeId::of::<T>(), setting_props);
+    }
+    fn add_itinerary(
```
k88hudson-cfa, line 236: I think we actually want this to add to 1 (and validate) so we can apply interventions more realistically (shifting time from one place to another place)

k88hudson-cfa (changes requested):

src/settings.rs
k88hudson-cfa: Actually, going to file a follow-up for this

k88hudson-cfa (approved):

Issue #67, opened 2025-03-10
Generalize `RateFnId` to refer to index of assignment for any natural history parameter


ChiragKumar9:
To model isolation guidance, we need to have person-specific natural history parameters (beyond just infectiousness rate functions, time to symptom onset, improvement, antigen positivity curves, etc.). We need to generalize `RateFnId` to refer to the index of any natural history parameter.

This issue may be scoped into two sub-issues as the exact work that needs to be done is figured out.

PR #75, opened 2025-03-21
Add an automated way of advancing through person properties. Closes #66.

### LINKED ISSUE #66: Add a clinical health status disease manager
Right now, we only track an individual's infectiousness/immune status (S, I, R) and have been intentional about ensuring these statuses are different from an individual's clinical health status (presymptomatic, asymptomatic, mildly symptomatic, severely symptomatic, dead). We need a simple manager that watches for a person becoming infected and then moves the person through these categories based on parameters (for now, these can be global parameters, but eventually we will want to add the option to make these person-specific parameters)

I don't think this manager should include hospitalization status because whether a severe infection becomes hospitalized may also be a function of policies, so I think that tracking should happen elsewhere.

See https://github.com/CDCgov/ixa-epi-isolation/tree/ckk_health_status for a very old template for a health status manager.

ChiragKumar9:
For our goal of modeling isolation guidance at the community level, we need some way of progressing people through individual-level (their particular) symptom progression. Concretely, we need some way of controlling when a person is presymptomatic, whether they have mild or moderate illness (we also technically need to include asymptomatics), when their symptoms have improved (according to our definition), etc.

There are a variety of solutions to this problem. At the simplest level, we could hardcode an event subscription that watches for the person becoming infected and then determines whether they should be presymptomatic or asymptomatic, if second check for presymptomatics that then schedules them to become mild or moderate (perhaps according to some probability), and a third check for either of those two cases to return them to health. This is very stringent and not flexible. It's also challenging to think about our use case in this framework -- we know the exact times at which a particular individual's symptoms change rather than abstract probabilities that define these events.

Instead, I define a trait that controls this sort of "progression" behavior between person properties -- effectively a Markov chain. Now, a user can define the trait on an empirical progression like we have for our isolation guidance case (I present the general form of this in the PR), or a probability distribution to stochastically determine someone's next event. Secondly, I introduce a way to make these progressions automated, so a user needs to only register a properties progression (and there can be multiple progressions per property -- one for each person, for instance) and people are moved through the progression automatically.

In this PR, I do not implement a way to read in an empirically-defined symptom progression from a file -- like we would need in our isolation guidance example -- although there is a way to implement this trivially. Additionally, I do not implement a way of accounting for correlations between the natural history parameters -- so a specific `RateFnId` determines an individual's property progression. Introducing this functionality is scoped in #67, but it is worth noting that there is a way to do it already. A user could implement the progression trait for a rate function struct, and then based on someone's rate function values, they have associated symptom progressions.

Finally, I think the ideas of automated progression in this PR are useful beyond just symptom progressions -- this idea is also useful for modeling policies that phases/people have to do different things based on times. The only aspect we would need to add in that framework would be some idea of a "trigger" which would set of the policy sequence, but that trigger does not necessarily need to be a person property change event (as it is implemented here).


RobertJacobsonCDC (approved):
This looks good to me.

Something to think about: Instead of enumerating all possible progressions, you can describe a state machine as a directed acyclic graph (DAG). You get a much more general mechanism without too much more work. On the other hand, its more abstract and possibly more difficult for modelers to understand—and in my opinion that's an important thing to consider.

src/symptom_progression.rs
confunguido, line 1: This file makes me think that we need more discussion about exactly how symptoms will be modeled here. I think the model should not have code that reads as example, unless it's in the tests. I think we are going to use symptom categories to model isolation guidance, so we should go with that.

src/symptom_progression.rs
```diff
@@ -0,0 +1,262 @@
...
+                Some(DiseaseSeverityValue::Recovered),
+                context.sample_distr(SymptomRng, Exp::new(1.0 / self.severe_time).unwrap()),
+            )),
+            Some(DiseaseSeverityValue::Recovered) | None => None,
+        }
+    }
+}
+
+pub fn init(context: &mut Context) -> Result<(), IxaError> {
+    // Todo(kzs9): We will read these progressions from a file from our isolation guidance modeling
```
confunguido, line 95: What do you mean here? what do these progressions look like in the file?

confunguido (changes requested):
The progression through symptom categories is looking good. I agree with Robert that I would prefer to describe these state progressions as a DAG. I think this is close enough to that so that we could iterate through it. 

The symptom module isn't quite there yet. We need to clearly describe those symptom transitions. Right now, it looks like there are options to do it in different ways.

src/symptom_progression.rs
ChiragKumar9: Reading in the progressions from a file should be a separate PR, so we can discuss the format there without bogging this down. When I made this PR, I filed #74 for us to remember to do that.

docs/clinical_health_status.md
```diff
@@ -0,0 +1,154 @@
...
+}
+
+define_person_property_with_default!(
+    IsolationGuidanceSymptom,
+    Option<IsolationGuidanceSymptomValue>,
+    None
+);
+```
+
+Isolation guidance symptoms are separate from an individual's actual disease severity. This is how
```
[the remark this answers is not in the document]
ChiragKumar9: Discussed this via Teams, but just so we have it in writing for posterity, I misinterpreted your request to be for another example of how one would use the progression manager but with a progression not defined empirically but probabilistically. I have corrected this in the review changes, but I kept an example in the tests.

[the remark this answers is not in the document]
ChiragKumar9: We had convergent evolution -- I see you made a bunch of these changes in [97bf84f](https://github.com/CDCgov/ixa-epi-isolation/commit/97bf84f1cee22ff657f4587a9fa32fa673f1bcdb) :)

src/symptom_progression.rs
```diff
@@ -0,0 +1,250 @@
...
+        };
+        context.register_property_progression(Symptoms, symptom_data);
+    }
+
+    event_subscriptions(context);
+}
+
+fn event_subscriptions(context: &mut Context) {
+    context.subscribe_to_event(
+        |context, event: PersonPropertyChangeEvent<InfectionStatus>| {
```
confunguido, line 100: Does everyone develop one of these symptoms? Do we need to define asymptomatic people or is that category 4? I say this because the proportion of people who follow the guidelines will be dictated by the amount of people developing symptoms.

confunguido (approved):
I think this is fine. Could you add a list of issues on the remaining tasks (e.g., read parameters from an input file)?

Issue #71, opened 2025-03-14
Bug in how infections are initialized


ChiragKumar9:
We assign infections by querying `initial_infections` number of people and setting them to infectious. We never check that the person has not been queried before. It is possible that the same person is picked twice, so we will have fewer observed infections in the initial population than we anticipated by seeding.

ChiragKumar9:
Gated by https://github.com/CDCgov/ixa/issues/301.

PR #80, opened 2025-03-24
Update Ixa

ChiragKumar9:
When switching to the version of Ixa on crates.io, I noticed that the stochastics changed. In particular, the bug referenced in #71 can have tests pass stochastically (hence we had not caught the problem), but now the stochastics have changed and the associated tests failed. I fixed the bug, but it's a curious observation that the stochastics changed.

I also noticed that we no longer needed `rand`, `paste`, and `ctor` manually in our `Cargo.toml` (which is what was causing the troubles that were fixed in https://github.com/CDCgov/ixa/pull/270), so potentially the two are related?

### DIFF
--- a/src/infection_propagation_loop.rs
+++ b/src/infection_propagation_loop.rs
@@ -44,11 +44,6 @@ fn schedule_recovery(context: &mut Context, person: PersonId) {
 /// - If `initial_infections` is greater than the population size.
 fn seed_infections(context: &mut Context, initial_infections: usize) -> Result<(), IxaError> {
     for _ in 0..initial_infections {
-        // Need to get the value of the derived person property first?
-        let _ = context.get_person_property(
-            context.sample_person(InfectionRng, ()).unwrap(),
-            InfectionStatus,
-        );
         let person = context.sample_person(
             InfectionRng,
             (InfectionStatus, InfectionStatusValue::Susceptible),
--- a/src/infection_propagation_loop.rs
+++ b/src/infection_propagation_loop.rs
@@ -40,12 +40,23 @@ fn schedule_recovery(context: &mut Context, person: PersonId) {
 }
 
 /// Seeds the initial population with a number of infectious people.
-fn seed_infections(context: &mut Context, initial_infections: usize) {
-    // First, seed an infectious population
-    for _ in 0..initial_infections {
-        let person = context.sample_person(InfectionRng, ()).unwrap();
-        context.infect_person(person, None);
+/// # Errors
+/// - If `initial_infections` is greater than the population size.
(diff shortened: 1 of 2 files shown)


confunguido (approved):
LGTM. I think we need to talk eventually about #85

KOVALW (changes requested):
LGTM, but I suggest using a query instead of a dummy count variable and `InfectionStatus` check below.

RobertJacobsonCDC (approved):
This all looks good to me. I was `clap` just a vestigial dependency that isn't being used?

ChiragKumar9:
> This all looks good to me. Was `clap` just a vestigial dependency that isn't being used?

Yep, `clap` was vestigial from the days of when we were using the custom runner!

KOVALW (approved):
Great - mentioning Issue [#301](https://github.com/CDCgov/ixa/issues/301) in ixa here as well to close

ChiragKumar9:
Merging this is gated by a resolution on https://github.com/CDCgov/ixa/issues/301. Either we stick with our workaround solution here (accessing the derived person property prior to initialization) or we need a fix in core Ixa.

RobertJacobsonCDC:
I'm working on CDCgov/ixa#301 now.

src/infection_propagation_loop.rs
```diff
@@ -40,12 +40,29 @@ fn schedule_recovery(context: &mut Context, person: PersonId) {
...
-fn seed_infections(context: &mut Context, initial_infections: usize) {
-    // First, seed an infectious population
+/// # Errors
+/// - If `initial_infections` is greater than the population size.
+fn seed_infections(context: &mut Context, initial_infections: usize) -> Result<(), IxaError> {
     for _ in 0..initial_infections {
-        let person = context.sample_person(InfectionRng, ()).unwrap();
-        context.infect_person(person, None);
+        // Need to get the value of the derived person property first?
+        let _ = context.get_person_property(
```
KOVALW, line 48: One extra comment is that I would probably move this initialization of the derived property type line (which will get removed once we merge in ixa PR 301) into the `init` function, with an appropriate comment tag that mentions this is to initialize the lazy loading of a derived property. Here it just looks like duplicate code until looking more closely

KOVALW (approved):
Still LGTM but wanted to flag the placement of the initialization line

src/infection_propagation_loop.rs
ChiragKumar9: This line of code was just added so the ixa dev team could see what we needed to do in order to get the querying to work. We will remove this line of code in prior to merging (because I'm holding on this PR until https://github.com/CDCgov/ixa/issues/301 is merged).

confunguido (approved):

PR #82, opened 2025-03-24
Create dependabot.yml. Closes #81.

### LINKED ISSUE #81: Add dependabot for automatic crate bumps
https://docs.github.com/en/code-security/getting-started/dependabot-quickstart-guide

We should also test that however we've set up dependabot it's working correctly.

ChiragKumar9:


ChiragKumar9:
@k88hudson-cfa thought you are most familiar with dependabot, but feel free to ask me to request someone else if easier!

RobertJacobsonCDC (commented):
This looks OK to me, but I don't have much experience with setting up Dependabot. The only comment I have is, in Ixa core we started using [Conventional Commits](https://www.conventionalcommits.org) to support automated releases, which means we use standard prefixes for commit messages. So you could have something like:

```yaml
version: 2
updates:
  - package-ecosystem: "cargo"
    directory: "/"
    schedule:
      interval: "weekly"
    versioning-strategy: "increase"
    commit-message:
      prefix: "fix: "
```

But it looks like you aren't using this convention in this repo, so a prefix is more a matter of personal preference.

.github/dependabot.yml
```diff
@@ -0,0 +1,18 @@
...
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+updates:
+  - package-ecosystem: "cargo" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "weekly"
+  # Enable version updates for GitHub Actions
+  - package-ecosystem: 'github-actions'
```
k88hudson-cfa, line 13: You should use a consistent quote style (also: is yaml not being linted? )

k88hudson-cfa (approved):
This looks good to me! R+ with minor nits on style

It would be nice to use this on core too (for that case I would add `prefix` so it fits with our conventional commit format)

.github/dependabot.yml
ChiragKumar9: It looks like yaml is being linted but this isn't being caught?? Odd

PR #83, opened 2025-03-25
included schools and workplaces in synthetic population script closes #77

### LINKED ISSUE #77: Use a synthetic population that is more representative of the settings we need

confunguido:
Added schools and workplaces to the script that generates the toy synthetic population. Modified population loader so that it gets that information in, but this will change once we incorporate the settings into the model.


ChiragKumar9 (changes requested):
I am not tied to my suggestions, and I am happy to have us file them as issues. I more just want us to discuss the person initialization pattern when the number of person properties being defined is unknown. I think this is a common pattern, and having a nice way of solving it will be helpful for the future.

KOVALW (approved):
lgtm

PR #84, opened 2025-03-25
Parse school and workplace ids

KOVALW:
**Updates to Rust code (`src/population_loader.rs`):**

* Added `schoolId` and `workplaceId` as optional fields to the `PeopleRecord` struct and corresponding property definitions.
* Updated the `create_person_from_record` function to parse and handle `schoolId` and `workplaceId` fields.
* Modified the test cases to include `schoolId` and `workplaceId` fields in the synthetic population data and added new tests to verify the correct handling of these fields.


confunguido (approved):
lgtm

PR #88, opened 2025-03-26
Use real rate functions from isolation guidance. Closes #86.

### LINKED ISSUE #86: Use real rate function data from isolation guidance
- [x] R script to make empirical rate function file in the right format of what we want ixa-epi from our Stan model output
- [x] Add `scale` parameter to empirical rate function that scales the hazard rates (which are not normalized by-person) to have epi-relevant meaning (parameter is calibrated)
- [x] What is the impact on the SIR curve?

ChiragKumar9:
- Adds some example data that mirrors the outputs we get from our Stan model of isolation guidance.
- Adds an R script that parses that data and turns it into an ixa-epi ready format (we keep that file in the repo to allow `cargo run` to use it and not fail on GitHub actions).
- Adds a `scale` parameter to the empirical rate functions that scales the relative hazard rates into an absolute rate of infection.

Also, here's what the SIR curve looks like when we use these new rate functions (scale = 1.0): 
image

Compared to what it looked like before (this is post addition of settings with each setting experiencing 25% of a person's infectiousness with alpha = 0.0):
image

We expect people to be more infectious since the rate of infection is higher with the triangle model with scale = 1.0 (the triangle's mean magnitude is about 7 and it's mean duration is about 15 days).


src/infectiousness_manager.rs
```diff
@@ -114,10 +118,13 @@ pub fn evaluate_forecast(
...
     let elapsed_t = context.get_elapsed_infection_time(person_id);
-    let current_infectiousness = total_rate_fn.rate(elapsed_t);
+    let current_infectiousness = match context.get_params().infectiousness_rate_fn {
+        RateFnType::EmpiricalFromFile { scale, .. } => total_rate_fn.rate(elapsed_t) * scale,
+        _ => total_rate_fn.rate(elapsed_t),
+    };
 
     assert!(
-        (f64::abs(current_infectiousness - forecasted_total_infectiousness) <= f64::EPSILON),
+        (f64::abs(current_infectiousness - forecasted_total_infectiousness) <= 2.0 * f64::EPSILON),
```
ChiragKumar9, line 127: This is where I've been running into the floating point arithmetic issues. I believe this is a numerical error. I have the change the float here to be `500.0 * f64::EPSILON` to get the model to run. Clearly, that's bad. I don't really know how to fix this. I noticed that the difference gets bigger more and more throughout an individual's infection -- I think that's because `ScaledRateFn.rate()` calls the base rate function but also multiplies by `self.scale` so the error just grows.

ChiragKumar9:
Marking as ready for review because the math, proposal, implementation, etc. are stabilized. But, this PR is currently gated. We have to fix the numerical errors in the use of the empirical rate functions (probably something in the quadratic inversion?). Second, we have to land #110 first because that proposes the file input scheme for a progression library. However, once we have merged that PR, all we need to do is specify the symptom progression file output from this PR in the input file for our ixa-epi model, so nothing will actually change code-wise.

scripts/generate_isolation_guidance_inputs.R
```diff
@@ -0,0 +1,144 @@
+## ===============================================#
+## Setup -----------------
+## ===============================================#
+library(tidyverse)
+library(sjmisc)
+source("scripts/functions.R")
+set.seed(108)
+
+# Jason's suggestion for basically adding a positive value (that we calibrate)
```
confunguido, line 9: Could you clean up this comment and just describe what it does?

confunguido (approved):
lgtm

confunguido:
It looks fine to me but I think cargo is failing.

ChiragKumar9:
> It looks fine to me but I think cargo is failing.

Yeah, I may ask for your help on this (if you want to go down a rabbit hole with me!)

scripts/generate_isolation_guidance_inputs.R
```diff
@@ -0,0 +1,169 @@
...
+  repeat {
+    x <- exp(rnorm(1, mean = log(mu), sd = sd))
+    if (x > minimum) {
+      return(x)
+    }
+  }
+}
+
+sampled_params <- sampled_params |>
+  dplyr::mutate(minimum_incubation_period = tp - wp) |>
```
eqmooring, line 51: I believe it would be more accurate to call this the `minimum_latent_period`.

scripts/generate_isolation_guidance_inputs.R
```diff
@@ -0,0 +1,169 @@
...
+  dplyr::mutate(
+    pre_infectiousness_period =
```
eqmooring, line 53: I may be misunderstanding, but would it be clearer to just call this the `latent_period`?

scripts/generate_isolation_guidance_inputs.R
```diff
@@ -0,0 +1,169 @@
...
+library(tidyverse)
+library(sjmisc)
+source("scripts/functions.R")
+set.seed(108)
+
+# Jason's suggestion: add some time before infectiousness starts
+# to get the time of infection. We can calibrate this time and use
+# it to check our assumption that infectiousness starts at logVL > 0.
+# The time we add is the disease latent period, and the latent period
+# plus the time to symptom onset is the incubation period.
```
eqmooring, line 13: Do you mean?

```suggestion
# plus the time from infectiousness onset to symptom onset is the incubation period.
```

scripts/generate_isolation_guidance_inputs.R
```diff
@@ -0,0 +1,169 @@
...
+  # treat them alone as the hazard rate.
+  # TODO (kzs9): allow for adding a viral_load_to_infectiousness function here
+  # which transforms the viral load to an infectiousness function using our
+  # isolation guidance assumptions for the different functional forms of
+  # infectiousness.
+  list(curve)
+}
+
+# For every parameter set, make a vector of the viral load over time trajectory.
+trajectories <- purrr::pmap_vec(
```
eqmooring, line 87: I think this is fine, but as discussed, it could be simplified to just three rows for each rate function. The times (I think) would be `c(tp - wp, wp, tp + wr) - infection_start_time` and the values would be `c(0, dp, 0)`

scripts/generate_isolation_guidance_inputs.R
```diff
@@ -0,0 +1,169 @@
...
+        pre_infectiousness_mean,
+        pre_infectiousness_std_dev
+      )
+  ) |>
+  # wp is the proliferation period -- time from logVL = 0 to peak logVL
+  # tp is the time of peak logVL wrt time since symptom onset
+  # so wp - tp is the time from logVL = 0 to symptom onset, and -(wp - tp)
+  # is the time at which infectiousness starts. We want to have the
+  # infection start before that, so we subtract some time.
+  dplyr::mutate(infection_start_time = tp - wp - pre_infectiousness_period)
```
eqmooring, line 66: Maybe do a `stopifnot()` to check that all `infection_start_time` values are negative?

scripts/generate_isolation_guidance_inputs.R
```diff
@@ -0,0 +1,169 @@
...
+# time since symptom. We need to convert times to be with respect to _infection_
+# onset. For this reason, we need to estimate the time of infection.
+# We do this by assuming that infection happens sometime before whenever we
+# assume infectiousness starts (logVL > 0), a calibratable amount of time.
+# Note that this time is the disease latent period, and the time from the start
+# of the infection to symptom onset is the incubation period
+# (this is -infection_start_time), so we can interpret these parameters
+# epidemiologically.
+# There core assumption here is really that infectiousness starts at logVL > 0.
+# If that is not true, we should see the pre infectiousness period run up
```
eqmooring, line 38: Minor comment: Consider revising documentation to make clear that calibration is not actually performed here.

scripts/generate_isolation_guidance_inputs.R
```diff
@@ -0,0 +1,169 @@
...
+  curve <- pmax(triangle_vl(
+    seq(infection_start_time, infection_start_time + max_time, dt),
+    dp, tp, wp, wr
+  ), 0)
+  # The triangle vl curves tell us the rate of transmission over time, so we
+  # treat them alone as the hazard rate.
+  # TODO (kzs9): allow for adding a viral_load_to_infectiousness function here
+  # which transforms the viral load to an infectiousness function using our
+  # isolation guidance assumptions for the different functional forms of
+  # infectiousness.
```
eqmooring, line 82: Yes, and to expand on this, it's important to be clear that the current code not only assumes that infectiousness scales with log viral load but that the critical baseline threshold is 0 log VL /mL.

scripts/functions.R
```diff
@@ -0,0 +1,49 @@
...
+
+  log_vl <- dplyr::if_else(t <= tp,
+    (dp / wp) * (t - (tp - wp)),
+    dp - (dp / wr) * (t - tp)
+  )
+
+  return(log_vl)
+}
+
+calculate_weibull_scale <- function(
```
eqmooring, line 25: The algebra looks correct to me, though I think we could think about (as an issue, not necessarily in this PR) tests that compare distribution of time to symptom improvement in this model versus what we simulated in stan.

scripts/generate_isolation_guidance_inputs.R
```diff
@@ -0,0 +1,169 @@
...
+library(sjmisc)
+source("scripts/functions.R")
+set.seed(108)
+
+# Jason's suggestion: add some time before infectiousness starts
+# to get the time of infection. We can calibrate this time and use
+# it to check our assumption that infectiousness starts at logVL > 0.
+# The time we add is the disease latent period, and the latent period
+# plus the time to symptom onset is the incubation period.
+pre_infectiousness_mean <- 0.5
```
eqmooring, line 14: Are these just semi-arbitrary choices? Just looking at the incubation periods in `input/library_symptom_parameters.csv`, some look extremely short while others look quite long. I wonder if you could turn down the standard deviation. But, as we've discussed, we never set out to estimate incubation periods in our analysis, so I see this issue as out-of-scope for now.

eqmooring (approved):
Thanks, Chirag. I am approving because I think the math / logic is fine. But I know I made a bunch of comments. Let me know if you want to discuss any of them.

scripts/generate_isolation_guidance_inputs.R
```diff
@@ -0,0 +1,163 @@
...
+    dt = dt
+  ), triangle_vl_wrt_infected_time,
+  .progress = TRUE
+)
+
+# Prepare the data for long format.
+# Turn the list of vectors into a matrix.
+trajectories <- as_tibble(trajectories, .name_repair = "minimal") |>
+  sjmisc::rotate_df()
+times <- seq(dt / 2, max_time, dt) # Add the times for each value
```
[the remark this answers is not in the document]
ChiragKumar9: Old code. No longer relevant now that we feed the triangles in directly rather than taking their hazard (which required a derivative, so the $dt/2$ was indicating that it was a midpoint derivative).

src/infectiousness_manager.rs
```diff
@@ -101,7 +102,10 @@ pub fn get_forecast(context: &Context, person_id: PersonId) -> Option<Forecast>
     let t = total_rate_fn.inverse_cum_rate(e)?;
 
     let next_time = context.get_current_time() + t;
-    let forecasted_total_infectiousness = total_rate_fn.rate(t);
+    let forecasted_total_infectiousness = match context.get_params().infectiousness_rate_fn {
```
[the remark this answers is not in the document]
ChiragKumar9: This is a great idea. Addressed in 2b37602.

scripts/generate_isolation_guidance_inputs.R
ChiragKumar9: Yep, for posterity, the reason we can simplify in this case is because our ixa-epi empirical rate function code linearly interpolates between specified rate function values, and the triangle viral load function is linear as it is. Because we will have to use the full timeseries when we have nonlinear hazard rates, for now I am keeping the explicit input structure.

Issue #76, opened 2025-03-21
Incorporate settings with the rest of the model


confunguido:

ChiragKumar9:
Closed by #93

PR #93, opened 2025-04-11
Add settings into model closes #46 and #76

### LINKED ISSUE #46: Forecast evaluation should take into account all settings
Right now when a forecast is evaluated, people are always assumed to be in a `Household`; we need to modify this to choose a random setting instead.

When people have actual itineraries this will be more complex but we could assume for now we could assume you're in any given setting for equal amounts of time.

confunguido:
Connected settings into the model in two ways: 
1) loading the synthetic population and assigning people to settings (school, work, etc)
2) Estimate total infectiousness from settings

To address:
- [ ] Settings does not have a notion of presence, so max and current total infectiousness are the same.
- [X] Ratios here are hard coded as well as alphas, which should come from some parameter file. 
- [ ] Need to test behavior of model with tests to make sure that transmission is happening in all settings as we expect

### DIFF
--- a/input/input.json
+++ b/input/input.json
@@ -4,6 +4,12 @@
     "seed": 123,
     "infectiousness_rate_fn": {"Constant": {"rate": 1.0, "duration": 5.0}},
     "initial_infections": 1,
+    "settings_properties": [
+      ["Home", 0.1],
+      ["Workplace", 0.1],
+      ["School", 0.1],
+      ["CensusTract", 0.1]
+    ],
     "report_period": 1.0,
     "synth_population_file": "input/people_test.csv"
   }
--- a/input/input.json
+++ b/input/input.json
@@ -10,7 +10,7 @@
       {"School": {"alpha": 0.1}},
       {"CensusTract": {"alpha": 0.1}}
     ],
-    "itinerary_write_fn": "SplitEvenly",
+    "itinerary_fn_type": "SplitEvenly",
     "report_period": 1.0,
     "synth_population_file": "input/people_test.csv"
   }
--- a/input/input.json
+++ b/input/input.json
@@ -10,6 +10,7 @@
       {"School": {"alpha": 0.1}},
       {"CensusTract": {"alpha": 0.1}}
     ],
+    "itinerary_write_fn": "SplitEvenly",
     "report_period": 1.0,
     "synth_population_file": "input/people_test.csv"
   }
--- a/input/input.json
+++ b/input/input.json
@@ -5,10 +5,10 @@
     "infectiousness_rate_fn": {"Constant": {"rate": 1.0, "duration": 5.0}},
     "initial_infections": 1,
     "settings_properties": [
-      ["Home", 0.1],
-      ["Workplace", 0.1],
-      ["School", 0.1],
-      ["CensusTract", 0.1]
+      {"Home": {"alpha": 0.1}},
+      {"Workplace": {"alpha": 0.1}},
+      {"School": {"alpha": 0.1}},
+      {"CensusTract": {"alpha": 0.1}}
     ],
     "report_period": 1.0,
     "synth_population_file": "input/people_test.csv"
--- a/src/contact.rs
+++ b/src/contact.rs
@@ -1,4 +1,3 @@
-use crate::settings::ContextSettingExt;
 use ixa::{define_rng, people::Query, Context, ContextPeopleExt, ContextRandomExt, PersonId};
 
 define_rng!(ContactRng);
@@ -8,16 +7,19 @@ pub trait ContextContactExt {
     /// properties, for example `(Alive, true)`.
     /// Returns None if there are no eligible contacts.
     #[allow(dead_code)]
-    fn get_contact<Q: Query>(&self, transmitter_id: PersonId, query: Q) -> Option<PersonId>;
-
-    /// Returns a potential contact from the transmitter give a pre-specified itinerary
-    /// Returns None if there aren't eligible contacts
-    // TODO: include a query
-    fn get_contact_from_settings(&self, transmitter_id: PersonId) -> Option<PersonId>;
+    fn get_population_contact<Q: Query>(
+        &self,
+        transmitter_id: PersonId,
+        query: Q,
+    ) -> Option<PersonId>;
 }
 
 impl ContextContactExt for Context {
-    fn get_contact<Q: Query>(&self, transmitter_id: PersonId, query: Q) -> Option<PersonId> {
+    fn get_population_contact<Q: Query>(
+        &self,
+        transmitter_id: PersonId,
+        query: Q,
+    ) -> Option<PersonId> {
         // Get list of eligible people given the provided query
         let possible_contacts = self.query_people(query);
         if possible_contacts.is_empty()
@@ -34,9 +36,6 @@ impl ContextContactExt for Context {
         }
         Some(contact_id)
     }
-    fn get_contact_from_settings(&self, transmitter_id: PersonId) -> Option<PersonId> {
-        self.draw_contact_from_itinerary(transmitter_id)
-    }
 }
 
 #[cfg(test)]
@@ -52,7 +51,7 @@ mod test {
         let mut context = Context::new();
         context.init_random(108);
         let transmitter = context.add_person((Alive, true)).unwrap();
-        let result = context.get_contact(transmitter, ());
+        let result = context.get_population_contact(transmitter, ());
         assert!(result.is_none());
     }
 
@@ -62,7 +61,7 @@ mod test {
         context.init_random(108);
         let transmitter = context.add_person((IsRunner, false)).unwrap();
         let contact = context.add_person((IsRunner, true)).unwrap();
-        let result = context.get_contact(transmitter, (IsRunner, true));
+        let result = context.get_population_contact(transmitter, (IsRunner, true));
         assert_eq!(result, Some(contact));
     }
 
@@ -74,7 +73,7 @@ mod test {
         context.add_person((Alive, false)).unwrap();
         context.add_person((Alive, false)).unwrap();
 
-        let observed_contact = context.get_contact(transmitter, (Alive, true));
+        let observed_contact = context.get_population_contact(transmitter, (Alive, true));
         assert!(observed_contact.is_none());
     }
 
@@ -89,7 +88,9 @@ mod test {
         context.add_person((Alive, false)).unwrap();
         context.add_person((Alive, false)).unwrap();
 
-        let observed_contact = context.get_contact(transmitter, (Alive, true)).unwrap();
+        let observed_contact = context
+            .get_population_contact(transmitter, (Alive, true))
+            .unwrap();
         assert_eq!(observed_contact, presumed_contact);
     }
 }
--- a/src/contact.rs
+++ b/src/contact.rs
@@ -1,4 +1,5 @@
 use ixa::{define_rng, people::Query, Context, ContextPeopleExt, ContextRandomExt, PersonId};
+use crate::settings::ContextSettingExt;
 
 define_rng!(ContactRng);
 
@@ -6,7 +7,13 @@ pub trait ContextContactExt {
     /// Returns a potential contact for the transmitter given a query of people
     /// properties, for example `(Alive, true)`.
     /// Returns None if there are no eligible contacts.
+    #[allow(dead_code)]
     fn get_contact<Q: Query>(&self, transmitter_id: PersonId, query: Q) -> Option<PersonId>;
+
+    /// Returns a potential contact from the transmitter give a pre-specified itinerary
+    /// Returns None if there aren't eligible contacts
+    // TODO: include a query
+    fn get_contact_from_settings(&self, transmitter_id: PersonId) -> Option<PersonId>;
 }
 
 impl ContextContactExt for Context {
@@ -26,6 +33,9 @@ impl ContextContactExt for Context {
                 possible_contacts[self.sample_range(ContactRng, 0..possible_contacts.len())];
         }
         Some(contact_id)
+    }    
+    fn get_contact_from_settings(&self, transmitter_id: PersonId) -> Option<PersonId> {
+        self.draw_contact_from_itinerary(transmitter_id)
     }
 }
 
--- a/src/contact.rs
+++ b/src/contact.rs
@@ -1,5 +1,5 @@
-use ixa::{define_rng, people::Query, Context, ContextPeopleExt, ContextRandomExt, PersonId};
 use crate::settings::ContextSettingExt;
+use ixa::{define_rng, people::Query, Context, ContextPeopleExt, ContextRandomExt, PersonId};
 
 define_rng!(ContactRng);
 
@@ -33,7 +33,7 @@ impl ContextContactExt for Context {
                 possible_contacts[self.sample_range(ContactRng, 0..possible_contacts.len())];
         }
         Some(contact_id)
-    }    
+    }
     fn get_contact_from_settings(&self, transmitter_id: PersonId) -> Option<PersonId> {
         self.draw_contact_from_itinerary(transmitter_id)
     }
--- a/src/infection_propagation_loop.rs
+++ b/src/infection_propagation_loop.rs
@@ -77,7 +77,7 @@ mod test {
 
     use ixa::{
         Context, ContextGlobalPropertiesExt, ContextPeopleExt, ContextRandomExt, ExecutionPhase,
-        IxaError, PersonPropertyChangeEvent,
+        PersonPropertyChangeEvent,
     };
     use statrs::{
         assert_almost_eq,
@@ -92,18 +92,14 @@ mod test {
             max_total_infectiousness_multiplier, InfectionContextExt, InfectionData,
             InfectionDataValue,
         },
-        parameters::{
-            ContextParametersExt, GlobalParams, Params, RateFnType, CORE_SETTING_VARIANTS,
-        },
+        parameters::{ContextParametersExt, CoreSettingsTypes, GlobalParams, Params, RateFnType},
         rate_fns::load_rate_fns,
-        settings::{
-            define_setting_type, ContextSettingExt, ItineraryEntry, SettingId, SettingProperties,
-        },
+        settings::{global_mixing_itinerary, ContextSettingExt, Global, SettingProperties},
     };
 
     use super::{schedule_recovery, seed_infections};
 
-    fn setup_context(seed: u64, rate: f64) -> Context {
+    fn setup_context(seed: u64, rate: f64, alpha: f64) -> Context {
         let mut context = Context::new();
         let parameters = Params {
             initial_infections: 3,
@@ -116,33 +112,23 @@ mod test {
             report_period: 1.0,
             synth_population_file: PathBuf::from("."),
             transmission_report_name: None,
-            settings_properties: [None; CORE_SETTING_VARIANTS],
+            settings_properties: vec![CoreSettingsTypes::Global { alpha }],
         };
         context.init_random(parameters.seed);
         context
             .set_global_property_value(GlobalParams, parameters)
             .unwrap();
-        context
-    }
-
-    define_setting_type!(Global);
-    fn global_mixing_itinerary(context: &mut Context, alpha: f64) -> Result<(), IxaError> {
         context.register_setting_type(Global {}, SettingProperties { alpha });
-
-        for i in context.query_people(()) {
-            let itinerary = vec![ItineraryEntry::new(&SettingId::<Global>::new(0), 1.0)];
-            context.add_itinerary(i, itinerary)?;
-        }
-        Ok(())
+        context
     }
 
     #[test]
     fn test_seed_infections() {
-        let mut context = setup_context(0, 1.0);
+        let mut context = setup_context(0, 1.0, 1.0);
         for _ in 0..10 {
             context.add_person(()).unwrap();
         }
-        global_mixing_itinerary(&mut context, 1.0).unwrap();
+
         load_rate_fns(&mut context).unwrap();
         seed_infections(&mut context, 5);
         let infectious_count = context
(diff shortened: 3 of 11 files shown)


ChiragKumar9 (changes requested):
It looks like commits from main were merged into Will's branch that was then merged into this branch (with a squash). This makes it look like changes previously merged into main are being suggested from this branch. Concretely, we have already updated the pre-commit yaml and Will already edited the assert statement in the infection propagation loop that checks for current infectiousness is always less than total infectiousness. Basically, a rebase and dropping the merge commits from Will's branch is needed. This may require manual resolution of merge conflicts since Will's branch was merged in with a squash rather than keeping commit history.

src/infectiousness_manager.rs
```diff
@@ -310,10 +320,10 @@ mod test {
...
         let mut context = setup_context();
         let p1 = context.add_person(()).unwrap();
-        assert_eq!(
-            max_total_infectiousness_multiplier(&context, p1),
-            TOTAL_INFECTIOUSNESS_MULTIPLIER
-        );
+
+        global_mixing_itinerary(&mut context, 1.0);
+
+        assert_eq!(max_total_infectiousness_multiplier(&context, p1), 0.0);
```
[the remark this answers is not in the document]
KOVALW: It's zero because there's only the infector present: $(1 -1)^\alpha$

src/infection_propagation_loop.rs
```diff
@@ -216,7 +232,7 @@ mod test {
         // attempt happens quickly, that increases the chance we see another in 1.0 time units, and
         // because there is basically this compensating relationship between the time and the number
         // of events, they "cancel" each other out to give a uniform distribution (handwavingly).
-        let num_sims: u64 = 10000;
+        let num_sims: u64 = 20_000;
```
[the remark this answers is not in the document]
KOVALW: The 10K replicates led to a KS statistic that was slightly above 0.1, as do 11K replicates. I've reduced to 15K replicates which consistently works across seeds. KS tests in general are on the more fickle end of divergence diagnostics.

src/infection_propagation_loop.rs
```diff
@@ -119,12 +123,22 @@ mod test {
         context
     }
 
+    define_setting_type!(Test);
+    fn global_mixing_itinerary(context: &mut Context, alpha: f64) {
+        for i in context.query_people(()) {
+            let itinerary = vec![ItineraryEntry::new(&SettingId::<Test>::new(0), 1.0)];
+            context.add_itinerary(i, itinerary).unwrap();
```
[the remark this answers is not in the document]
KOVALW: Still going to unwrap in the tests as they should panic if add_itinerary fails

src/infection_propagation_loop.rs
```diff
@@ -237,6 +253,9 @@ mod test {
             load_rate_fns(&mut context).unwrap();
             // Add our infectious fellow.
             let infectious_person = context.add_person(()).unwrap();
+
+            global_mixing_itinerary(&mut context, 1.0);
+
             context.infect_person(infectious_person, None);
             // Get the total infectiousness multiplier for comparison to total number of infections.
             if total_infectiousness_multiplier.is_none() {
```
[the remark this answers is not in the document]
KOVALW: The max infectiousness multiplier for a person is now the number of people in their setting multiplied by the scaling factor, so the default is no longer 2.0, leading to a different total rate based on number of people in the setting and therefore various failures for when there is only one other person in a setting. Doubling the rate to 3.0 (effectively including the previous max_multiplier of 2.0) allows a 10K replicate test to pass. Rather than depend on some scaling constant, I wanted the test to pass with the true max_multiplier equivalent to one other individual in the transmitter's setting, the original simulation, and keeping the rate the same.

src/settings.rs
```diff
@@ -335,6 +340,32 @@ impl ContextSettingExt for Context {
     }
 }
 
+pub fn init(context: &mut Context) {
+    let Params {
+        settings_properties,
+        ..
+    } = context.get_params();
+
+    let settings_properties = settings_properties.clone();
```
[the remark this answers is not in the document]
KOVALW: Vec doesn't implement Copy, I've opted to declare a `CORE_SETTING_VARIANTS` variable that makes a fixed array that's copyable. This is also easier to run through `validate()`

src/infection_propagation_loop.rs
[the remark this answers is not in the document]
ChiragKumar9: Right, I agree. But what I'm wondering is what changed in the underlying code that forced us to need to increase the number of simulations we run for this test? I don't conceptually see what changed: in both cases, we had homogenous mixing, we still do now, the only thing that changed is the alpha value? I' m just trying to trace down why this change is necessary.

src/infection_propagation_loop.rs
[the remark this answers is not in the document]
ChiragKumar9: Ah, this is the answer to my question up above.

src/infection_propagation_loop.rs
```diff
@@ -119,12 +123,22 @@ mod test {
         context
     }
 
+    define_setting_type!(Test);
+    fn global_mixing_itinerary(context: &mut Context, alpha: f64) {
+        for i in context.query_people(()) {
+            let itinerary = vec![ItineraryEntry::new(&SettingId::<Test>::new(0), 1.0)];
```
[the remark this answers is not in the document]
KOVALW: Now contains parameter values that are labelled - this is a more general issue though. Issue created #109

src/settings.rs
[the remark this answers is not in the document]
KOVALW: This proved to be more cumbersome in practice than it was to code up. I've reverted to `vec![]` so we can optionally include or exclude settings (such as `Global`) more easily.

src/settings.rs
```diff
@@ -341,18 +407,78 @@ impl ContextSettingExt for Context {
...
+    for i in context.query_people(q) {
+        let itinerary = vec![ItineraryEntry::new(
+            &SettingId::<Global>::new(0),
+            mixing_time_proportion,
+        )];
+        context.add_itinerary(i, itinerary)?;
+    }
+    let query = q.get_query();
+    if query.is_empty() {
+        context.subscribe_to_event::<PersonCreatedEvent>(move |context, event| {
```
ChiragKumar9, line 436: Why not have people global mix if they are created later and satisfy the query, even if the query currently has people in it?

src/settings.rs
```diff
@@ -341,18 +407,78 @@ impl ContextSettingExt for Context {
...
+            }
+            CoreSettingsTypes::School { alpha } => {
+                context.register_setting_type(School {}, SettingProperties { alpha: *alpha });
+            }
+            CoreSettingsTypes::Workplace { alpha } => {
+                context.register_setting_type(Workplace {}, SettingProperties { alpha: *alpha });
+            }
+            CoreSettingsTypes::Global { alpha } => {
+                context.register_setting_type(Global {}, SettingProperties { alpha: *alpha });
+                global_mixing_itinerary(context, 1.0, ())?;
```
ChiragKumar9, line 470: Should alpha here always be called with 1.0, or should it be called with the alpha specified in the setting type?

src/settings.rs
```diff
@@ -341,18 +407,78 @@ impl ContextSettingExt for Context {
...
+}
+
+pub fn init(context: &mut Context) -> Result<(), IxaError> {
+    let Params {
+        settings_properties,
+        ..
+    } = context.get_params();
+
+    for setting in &settings_properties.clone() {
+        match setting {
```
ChiragKumar9, line 455: This is also another use case for how we can clean up the settings ergonomics -- I think that putting the setting data in a `RefCell` will make it so that we don't need explicit registration like this...

src/settings.rs
```diff
@@ -56,6 +60,39 @@ impl ItineraryEntry {
     }
 }
 
+type ItineraryEntryWriter<T> = dyn Fn(&Context, &SettingId<T>) -> Box<ItineraryEntry>;
```
ChiragKumar9, line 63: Why do you need to box the itinerary entry?

src/settings.rs
```diff
@@ -155,42 +193,52 @@ pub trait ContextSettingExt {
...
-    fn get_contact_internal(
+    fn get_contact_internal<T: Query>(
         &self,
         person_id: PersonId,
         setting_type: TypeId,
         setting_id: usize,
+        q: T,
     ) -> Option<PersonId> {
         if let Some(members) = self.get_setting_members_internal(setting_type, setting_id) {
+            let query_population = self.query_people(q);
```
ChiragKumar9, line 234: FWIW, I would move this line to be below the early return to save you a querying if members only contains one person

src/settings.rs
```diff
@@ -155,42 +193,52 @@ pub trait ContextSettingExt {
...
             if members.len() == 1 {
                 return None;
             }
+            let members = inner_join(members, &query_population);
             let mut contact_id = person_id;
             while contact_id == person_id {
-                contact_id = members[self.sample_range(SettingsRng, 0..members.len())];
+                contact_id = *members[self.sample_range(SettingsRng, 0..members.len())];
```
ChiragKumar9, line 241: Is there a way to use `QueryAnd` here from core ixa to basically join these two queries and then just sample a person from the joint query? Effectively only ever using core ixa methods.

src/settings.rs
```diff
@@ -155,42 +193,52 @@ pub trait ContextSettingExt {
...
         &self,
         setting_type: TypeId,
         setting_id: usize,
     ) -> Option<&Vec<PersonId>>;
+    fn get_itinerary_write_rules<T: SettingType + 'static>(&self) -> Box<ItineraryEntryWriter<T>>;
 }
 
 impl ContextSettingInternalExt for Context {
-    fn get_contact_internal(
+    fn get_contact_internal<T: Query>(
```
ChiragKumar9, line 226: So I looked over this code with adding a query to the contact sampling. I really like the idea. But, I don't love the ergonomics right now. I also think it's a bit too much for this PR to be able to properly provide feedback. Could we pull it out and put it in a separate PR?

src/settings.rs
```diff
@@ -568,16 +694,16 @@ mod test {
         assert_eq!(
             person_b,
             context
-                .get_contact::<Home>(person_a, SettingId::<Home>::new(0))
+                .get_contact::<Home, _>(person_a, SettingId::<Home>::new(0), ())
                 .unwrap()
         );
         assert!(context
-            .get_contact::<CensusTract>(person_a, SettingId::<CensusTract>::new(0))
+            .get_contact::<CensusTract, _>(person_a, SettingId::<CensusTract>::new(0), ())
```
ChiragKumar9, line 701: I think you can actually just drop the turbofish here entirely. Actually, if that works, it may largely alleviate my ergonomic concerns from above.

src/settings.rs
```diff
@@ -626,12 +752,84 @@ mod test {
                 .clone();
 
             let _ = context.add_itinerary(person, itinerary_home);
-            let contact_id_home = context.draw_contact_from_itinerary(person);
+            let contact_id_home = context.draw_contact_from_transmitter_itinerary(person, ());
```
ChiragKumar9, line 755: From an implementation perspective, I think we should expose two functions to the user -- one where we just draw a contact from a transmitter itinerary, and one where we draw with a query. I think these should be two separate functions (the blanket draw from itinerary internally calls the second function with a null query), but this would make usability better since drawing with a null query is such a common use case.

src/settings.rs
```diff
@@ -626,12 +752,84 @@ mod test {
...
+        - Create 7th person with itinerary at home and census tract
+        - Assign Age property to people and query for only Age = 42
+        - Call "draw contact from itinerary":
+          + Compute total infectiousness
+          + Draw a setting weighted by total infectiousness
+          + Sample contact from chosen setting
+         - Test 1 Itinerary with 0 proportion at census tract, contacts drawn should be from home (0-2)
+         - Test 2 Itinerary with 0 proportion at home, contacts should be drawn from census tract (3-6)
+         */
+        for seed in 0..100 {
```
ChiragKumar9, line 780: It's not clear to me at this point in the code development why we need to run this test 100 times?

ChiragKumar9: Are you just trying to show that the code works always, and not just because of a fluke of the random seed?

ChiragKumar9 (changes requested):
Some really nice progress. There's a lot here (and  much of it is code we will use a lot), so I'd like to be able to take a look at this again before we land it.

input/input.json
```diff
@@ -4,6 +4,13 @@
     "seed": 123,
     "infectiousness_rate_fn": {"Constant": {"rate": 1.0, "duration": 5.0}},
     "initial_infections": 1,
+    "settings_properties": [
+      {"Home": {"alpha": 0.1}},
+      {"Workplace": {"alpha": 0.1}},
+      {"School": {"alpha": 0.1}},
+      {"CensusTract": {"alpha": 0.1}}
+    ],
+    "itinerary_write_fn": "SplitEvenly",
```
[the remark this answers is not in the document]
ChiragKumar9: Also, for posterity, since we are introducing the use of settings in this PR, could we get a figure in the GitHub page that shows how the SIR dynamics change now that we've added settings? Acknowledging that our rate functions right now are.... arbitrary

src/contact.rs
```diff
@@ -6,11 +6,20 @@ pub trait ContextContactExt {
     /// Returns a potential contact for the transmitter given a query of people
     /// properties, for example `(Alive, true)`.
     /// Returns None if there are no eligible contacts.
-    fn get_contact<Q: Query>(&self, transmitter_id: PersonId, query: Q) -> Option<PersonId>;
+    #[allow(dead_code)]
+    fn get_population_contact<Q: Query>(
```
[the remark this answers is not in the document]
KOVALW: @confunguido I agree that `contact.rs` could be cut since it's not offering additional functionality we use. Thoughts?

src/settings.rs
KOVALW: Opted for `match_person` since we can't query `PersonId`s, greatly simplifies things to remove the `inner_join` function

src/settings.rs
KOVALW: How frequently are we going to query with a null when selecting amongst contacts who should be susceptible?

src/settings.rs
KOVALW: You're right - there's only one possible contact to draw per query so removing the for loop

src/parameters.rs
```diff
@@ -10,6 +9,21 @@ pub enum RateFnType {
...
+pub enum CoreSettingsTypes {
+    Home { alpha: f64 },
+    School { alpha: f64 },
+    Workplace { alpha: f64 },
+    CensusTract { alpha: f64 },
+    Global { alpha: f64 },
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone, Copy)]
+pub enum ItineraryWriteFnType {
```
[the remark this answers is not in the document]
KOVALW: Ya I had envisioned this as eventually including "read from file" or allowing a user to specify joint distributions for time in each setting, which in my opinion would look like adefining a new `type` of `Fn` that returns a `Box<Itinerary>`

src/settings.rs
KOVALW: Good catch - the `TypeId` of the trait doesn't need to be `Box`'d, so I've removed it

src/settings.rs
KOVALW: The 1.0 here is to specify itinerary mixing time proportion, not alpha. Added a comment to clarify

src/settings.rs
KOVALW: Changed to get rid of the check and simply pass in the query with a `match_person` check

src/settings.rs
```diff
@@ -341,18 +407,78 @@ impl ContextSettingExt for Context {
...
             None
         }
     }
 }
 
+/// Function to assign a Global itinerary item to all individuakls in a query
+/// If the query is empty, future people will be added by subscribing to the `PersonCreatedEvent`
+/// Function must be called after people are added if the simulation is not running and able to listen
+/// for events. People added after global itinerary is set but before `conterxt.execute()` will not have the global itinerary
+pub fn global_mixing_itinerary<Q: Query>(
```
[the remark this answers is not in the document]
KOVALW: This is a great point, I've added a comment to the documentation for now but we need to think about how to listen for changes to person properties that might make a person switch off the applicable query

src/population_loader.rs
```diff
@@ -20,39 +23,33 @@ pub struct PeopleRecord<'a> {
...
+    }
+    if !school_string.is_empty() {
+        itinerary_person.add_setting_to_itinerary::<School>(context, school_string.parse()?)?;
+    }
+
+    // Add the settings consistent across all individuals (Home, CensusTract) and rescale to 1.0
+    itinerary_person
+        .add_setting_to_itinerary::<Home>(context, home_id.parse()?)?
+        .add_setting_to_itinerary::<CensusTract>(context, tract.parse()?)?
+        .rescale();
```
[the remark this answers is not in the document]
KOVALW: We don't - removing for now

src/parameters.rs
```diff
@@ -40,6 +58,47 @@ fn validate_inputs(parameters: &Params) -> Result<(), IxaError> {
             "The report writing period must be non-negative.".to_string(),
         ));
     }
+
+    let mut variant_counts = [0; CORE_SETTING_VARIANTS];
+    for setting in &parameters.settings_properties {
+        let alpha_value;
+        match setting {
```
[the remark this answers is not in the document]
KOVALW: I've gone with just a HashMap check of Id to value, calling `AsRef` to get the alpha values out. This is a more clear compile time failure if the `AsRef` match isn't updated.

src/parameters.rs
```diff
@@ -10,6 +14,44 @@ pub enum RateFnType {
...
+#[derive(Debug, Clone, Copy, Deserialize, Serialize)]
+pub enum CoreSettingsTypes {
+    Home { alpha: f64 },
+    School { alpha: f64 },
+    Workplace { alpha: f64 },
+    CensusTract { alpha: f64 },
+    Global { alpha: f64 },
+}
+
+impl AsRef<f64> for CoreSettingsTypes {
```
ChiragKumar9, line 26: Love the idea, I would suggest implementing your own method like `get_alpha()` on the type rather than hijacking a method in a well-known trait that has a use case

src/parameters.rs
```diff
@@ -10,6 +14,44 @@ pub enum RateFnType {
...
+            CoreSettingsTypes::Home { alpha }
+            | CoreSettingsTypes::School { alpha }
+            | CoreSettingsTypes::Workplace { alpha }
+            | CoreSettingsTypes::CensusTract { alpha }
+            | CoreSettingsTypes::Global { alpha } => alpha,
+        }
+    }
+}
+
+impl Display for CoreSettingsTypes {
```
ChiragKumar9, line 38: Ugh, I see why you have to do this, but it's clunky. The macro as suggested here https://github.com/CDCgov/ixa/issues/220 would be really useful for you too!

src/parameters.rs
```diff
@@ -40,6 +86,23 @@ fn validate_inputs(parameters: &Params) -> Result<(), IxaError> {
...
+
+    for setting in &parameters.settings_properties {
+        let setting_type = setting.to_string();
+        let alpha = setting.as_ref();
+        if *alpha < 0.0 {
+            return Err(IxaError::IxaError(
+                "The alpha values for each setting must be non-negative.".to_string(),
+            ));
+        }
+        if settings_map.insert(setting_type, *alpha).is_some() {
```
ChiragKumar9, line 100: Beautiful. And it returns early if there's a duplicate rather than waiting to the end!

src/parameters.rs
```diff
@@ -40,6 +86,23 @@ fn validate_inputs(parameters: &Params) -> Result<(), IxaError> {
             "The report writing period must be non-negative.".to_string(),
         ));
     }
+
+    let mut settings_map: HashMap<String, f64> = HashMap::new();
+
+    for setting in &parameters.settings_properties {
+        let setting_type = setting.to_string();
```
ChiragKumar9, line 93: Nit: would call this `setting_string` or something because `setting_type` has a meaning

ChiragKumar9: (though, yes, I agree this is the setting type strictly speaking)

src/settings.rs
```diff
@@ -56,6 +61,46 @@ impl ItineraryEntry {
...
+    ) -> Result<&mut Self, IxaError> {
+        let setting_id = &SettingId::<T>::new(setting_id_value);
+        let writer = context.get_itinerary_write_rules::<T>();
+        self.push(writer(context, setting_id));
+        Ok(self)
+    }
+    /// Merge the current itinerary with another itinerary
+    /// Takes in an existing itinerary vector and filters for entries that have no matches in the provided vector
+    /// Other values are ignored as they are already present in the input self vector
+    fn merge_itinerary(&mut self, other_itinerary: &Self) {
```
ChiragKumar9, line 90: This is so much better than the inner join

src/settings.rs
```diff
@@ -341,18 +475,77 @@ impl ContextSettingExt for Context {
...
         } else {
             None
         }
     }
 }
 
+/// Function to assign a Global itinerary item to all individuakls in a query
+/// If the query is empty, future people will be added by subscribing to the `PersonCreatedEvent`
+/// Function must be called after people are added if the simulation is not running and able to listen
+/// for events. People added after global itinerary is set but before `conterxt.execute()` will not have the global itinerary
```
ChiragKumar9, line 489:
```suggestion
/// for events. People added after global itinerary is set but before `context.execute()` will not have the global itinerary
```

ChiragKumar9 (approved):
I think this looks good! Thanks for working through this. Some small things, but I don't need to review it again. That being said, I think we need to file some things that come up in this PR as issues -- two big ones being auditing the use of itineraries in terms of how we end up actually using them (you have some of this filed in #97 and associated issues), and how to handle new people getting itineraries.

src/parameters.rs
KOVALW: Definitely agreed on this

src/settings.rs
KOVALW: This got outdated somehow, but I've made the change

KOVALW (approved):
@confunguido this gets at the functionality and extracting out parameters, which I think is sufficient for the PR. It doesn't address the issues on current vs. max infectiousness (which I think will be a separate issue) and validation tests (#112).

KOVALW:
Last two commits temporarily resolve the error brought up in Issue #115 but discussion should still be had about when we make the evaluation on Susceptible for infection attempts (using a query to incorporate this information into `current_total_infectiousness` or keeping the `infection_attempt` `match` statement

PR #95, opened 2025-04-15
Set R version in pre-commit to 4.4

ChiragKumar9:
Pre-commit is failing on open PRs. This is because Github actions is grabbing R 4.5, and for some reason this is causing an issue with getting the package `digest` installed in pre-commit. Reverting back to R 4.4 solves the problem. This is a documented issue on other CFA CDCGov repos.


KOVALW (approved):
Thanks!

PR #106, opened 2025-04-21
Allow relationships between natural history parameters (for isolation guidance, rate functions and symptom progressions). Closes #67.

### LINKED ISSUE #67: Generalize `RateFnId` to refer to index of assignment for any natural history parameter

ChiragKumar9:
This PR introduces a natural history parameter manager that enables modeling correlations and relationships between libraries of natural history parameters.

Concretely, to model isolation guidance outputs from the Stan model, we have a library of empirical rate functions -- that is, we have a set of possible hazard of infection curves, and we assign a person a curve when they are infected. We repeatedly query this particular curve throughout their infection. We also have a set of symptom improvement time distributions -- but the symptom improvement time distributions are directly related to the rate functions so that each rate function is associated with a particular symptom improvement distribution. In other words, in our isolation guidance case, the rate function id should equal the idea of the symptom improvement time distribution used for this person.

This example is emblematic of the broader problem of needing to enforce a relationship/correlation between libraries of natural history parameters. Broadly, the problem statement is that we want to have some way of specifying the index from a library of natural history parameters assigned to a person. We want to allow for the index chosen to be a function of other things, such as the index for a different natural history parameter or person properties or random assignment, and once an index for a parameter is set for a person, we want to be able to reproducibly query it (that way we can query the person's infectiousness rate function when they are infecting someone else, not have to carry around the assigned rate function id, and then be able to query it the next time they are infecting someone and get the same id out).

Thinking about the problem in this generic fashion gives us an extra benefit of popping out some other functionality -- by providing an interface that lets us specify natural history ids based on `&Context` and `PersonId`, we have effectively also provided the tooling necessary to pre-assign people specific infectiousness rate functions (let's say that rate functions depend on some demographic characteristics or we are modeling a very real-world case where we know that person 1's rate function was x).

Ultimately, this PR does three things:

1. Introduce an interface for registering a way to assign natural history ids and then lazily assign and the id for a particular parameter and person.
2. Apply this interface to the specific example of isolation guidance where we know that the id of the rate function should be the id of the symptom improvement time distribution.
3. Refactor the `RateFnId(usize)` and manual querying and assignment of rate functions in the infection propagation loop to use the new `NaturalHistoryId` infrastructure. The new method lazily assigns ids, improving code efficiency.

This proposal is not without problems. Here are the two greatest issues I ran into when playing with the code.

1. This code depends on types that implement the `NaturalHistoryParameter` trait, but there's not an obvious choice of what types should implement this trait. In the `property_progression_manager`, I provide a blanket implementation for all types that implement `PersonProperty` because person properties are effectively being treated as natural history parameters in that code, but this isn't the clear choice throughout. For the rate functions, I wanted to make the struct that we use to refer to the rate function data container `struct RateFnPlugin` implement the `NaturalHistoryParameter` trait, but that struct is private. Instead, I made a new public unit struct called `RateFn` and implemented the trait for that -- because it is a public struct, I can use it elsewhere (such as in the code for the symptom progression manager where I enforce the relationship between natural history parameter ids we need for isolation guidance)
2. The lazy assignment/querying structure takes an immutable reference to the data container where the ids are stored. However, this requires that at least one mutable call to the data container has been made previously to ensure the data container is initialized. There doesn't seem to be a way around this. `mod random` in core ixa uses the same structure and power of `RefCell<>` and has to do the same thing.


src/natural_history_parameter_manager.rs
```diff
@@ -0,0 +1,328 @@
...
+        if let Some(potential_id) = container
+            .ids
+            .borrow()
+            .get(&TypeId::of::<T>())
+            .and_then(|ids| ids.get(&person_id))
+        {
+            return *potential_id;
+        }
+
+        // Is there an assignment function for this parameter?
```
KOVALW, line 135: It might be more clear to have an if else block for storing the id here, meaning to stop at `let id = assigner(...).id()` and have an else that FIRST registers a default "random" type assigner, so that we're still tracking assigners of traits which are registered by querying, and then assigns id using the new random assigner. Then after the if else block, run the `container.ids...insert()` and return `id`

src/rate_fns/rate_fn_storage.rs
```diff
@@ -68,6 +73,12 @@ pub fn load_rate_fns(context: &mut Context) -> Result<(), IxaError> {
             add_rate_fns_from_file(context, file)?;
         }
     }
+
+    context.register_parameter_id_assignment(RateFn, |context, _| {
```
KOVALW, line 77: How is this different from the default random assignment sourced from library size?

src/rate_fns/rate_fn_storage.rs
```diff
@@ -1,57 +1,62 @@
...
-pub struct RateFnId(usize);
+define_rng!(InfectiousnessRng);
 
 struct RateFnContainer {
     rates: Vec<Box<dyn InfectiousnessRateFn>>,
 }
 
+pub struct RateFn;
+
+impl NaturalHistoryParameter for RateFn {
```
KOVALW, line 23: What happens if the library size is dynamic, as might be the case here?

KOVALW (commented):
Love the implementation, which feels natural. I think a specified default assigner rather than an implicit method would be preferable but it's not a strong need

src/natural_history_parameter_manager.rs
ChiragKumar9: I love this idea from an internal coding implementation perspective, but it would require either (a) that we have a mutable reference to context in `get_parameter_id` so that we can just call `register_parameter_id_assigner` or (b) that we put the assignment function hashmap in a RefCell. I tried (a), and it basically makes this manager unusable unfortunately -- we tend to need `get_parameter_id` in places where we already have taken or need other mutable references to context. I'm fine with (b), but I want to do a profiling to see what the impact of that RefCell addition is -- what I read in the books generally tells me RefCells are slow/bad?

Or maybe our more dev friends can help us here :)

src/rate_fns/rate_fn_storage.rs
ChiragKumar9: Yep, good eye -- it's not different! And this is one of the downsides of this manager -- you need to make at least one call to `register_parameter_id_assigner` prior to being able to call `get_parameter_id`. This is because the current ixa structure requires us to call `get_data_container_mut` prior to having a data container default initialized or being able to return `Some(data_container)` when calling `context.get_data_container` not mutable version (this is why, for instance, you have to call `context.init_random(seed)` prior to drawing a random number). Both @RobertJacobsonCDC and I think this is annoying, so we're thinking about a way to store the data containers in a RefCell that would enable you to get the default data container if calling `context.get_data_container` prior to any initialization call to the data container.

src/rate_fns/rate_fn_storage.rs
ChiragKumar9: An excellent question. The assignment ids would span the whole range of the natural history parameter library size, unless the assignment function constrained them. Note that at this time the assignment function output is never checked to make sure that it is less than the library size. Not sure if that's a good idea, and definitely fixable, but worth stating up front as a caveat.

src/natural_history_parameter_manager.rs
```diff
@@ -0,0 +1,328 @@
...
+
+/// Specifies behavior for obtaining a natural history id from a type.
+pub trait NaturalHistoryId {
+    /// Returns the index of a natural history parameter in a library of parameters.
+    fn id(&self) -> usize;
+}
+
+// By default, natural history indeces will be usize. This is a convenience implementation for
+// common user code that returns a usize id as the natural history id.
+impl NaturalHistoryId for usize {
```
RobertJacobsonCDC, line 25: In my opinion, I think we should either make a [[newtype idiom](https://doc.rust-lang.org/rust-by-example/generics/new_types.html)](https://doc.rust-lang.org/rust-by-example/generics/new_types.html), or just use a `usize` directly.

The newtype pattern is when you wrap an existing type in a tuple‐struct (or single‐field struct) to create a distinct, zero‐cost type:

```rust
pub struct NaturalHistoryId(usize);
```

Advantages of a newtype over the raw type:

- Stronger type safety
- Custom trait impls: You’re free to implement—or forbid—traits on the newtype even if the inner type already implements them (helpful with the orphan rules).
- Encapsulation: You can hide the inner type (“private field”) to enforce  invariants (e.g. always positive, non‐zero). [[This can be surprisingly powerful.](https://plv.mpi-sws.org/rustbelt/ghostcell/)](https://plv.mpi-sws.org/rustbelt/ghostcell/)
- Zero runtime cost: has the same internal representation as the wrapped type.

Cons of using a newtype:

- Boilerplate: You often need to impl or derive traits (Deref, From, Into, Display, etc.) to make the newtype ergonomic.
- Interoperability friction: Every time you need the inner value, you must unpack it (e.g. .0, .into(), or via a conversion method). When you find yourself implementing `From<innertype>` (and related), chances are using a newtype is just adding complexity without giving any advantage!
- Cluttered API surface: Client code needs to know about yet another thing.



I would use a trait only in those cases where a single instance of the interface needs to be able to take multiple different types in a generic way at runtime. But if I understand this API correctly, I think the idea is that the type is actually fixed at compile time, and we sort of need it to act like a `usize`.

RobertJacobsonCDC (changes requested):
I think this is a really reasonable API. Regarding the philosophical question of having another layer of abstraction (this API) versus having multiple ad hoc implementations, my opinion is that this API brings a lot to the table without really imposing any downsides. 

One disadvantage, if you can call it a disadvantage, is that it takes intellectual effort to evaluate how a potential application might or might now fit into this particular framework. But in my personal opinion, that intellectual effort is effort that should be spent anyway. 

I feel like the API doesn't commit you to a whole lot, so if it turns out to be the wrong abstraction, a refactor shouldn't be too painful. 

If there are more ways to connect the abstract interface to concrete examples, that would really be helpful. But you have a chicken-and-egg problem if you need to infrastructure to exist before you start using it, and you need to have uses of it in order to justify the existence of the interface. So in this case I would err on the side of adoption. It's pretty low stakes.

src/natural_history_parameter_manager.rs
```diff
@@ -0,0 +1,328 @@
...
+        *self
+    }
+}
+
+type IdAssigner<I> = dyn Fn(&Context, PersonId) -> I;
+
+#[derive(Default)]
+struct NaturalHistoryParameterContainer {
+    parameter_id_assigners: HashMap<TypeId, Box<IdAssigner<Box<dyn NaturalHistoryId>>>>,
+    ids: RefCell<HashMap<TypeId, HashMap<PersonId, usize>>>,
```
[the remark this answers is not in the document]
ChiragKumar9: Agreed on the vector front, but what would I index the vector by? I don't have the person ids available outside of the ixa crate (not even outside of the people module?).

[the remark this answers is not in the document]
rljacobson: Oh snap, you're right! Let me think about this some more. I've started a set of notes about our `DataPlugin` API that I'll add this to, also.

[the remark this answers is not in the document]
ChiragKumar9: I actually hope that some of the functionality described in this PR and this use case could serve as a motivation for how to extend person properties. I think a lot of the functionality described here could be grouped into core person properties, but that's a refactoring question for when discuss what in our ixa-epi-isolation managers should be made into an ixa-epi library, and what should go into core ixa!

src/natural_history_parameter_manager.rs
ChiragKumar9: Deciding to just use `usize` directly as I can't immediately think of use cases where we would want the trait functionality and then calculating the id on the fly!

src/natural_history_parameter_manager.rs
```diff
@@ -0,0 +1,370 @@
...
+            .register_parameter_id_assigner(CulturePositivity, |context, person_id| {
+                context.get_parameter_id(TestingPatterns, person_id)
+            })
+            .unwrap();
+        let culture_id = context.get_parameter_id(CulturePositivity, person);
+        let testing_id = context.get_parameter_id(TestingPatterns, person);
+        assert_eq!(culture_id, testing_id);
+    }
+
+    #[test]
```
confunguido, line 326: It would be good to add another test where there are some ids missing from one file to the other. I think the model should panic if these ids aren't contiguous or start in 1.

confunguido (changes requested):
Just add the input file validation step. It looks good otherwise.

src/rate_fns/rate_fn_storage.rs
confunguido, line 1: If I modify the file `two_rate_fns.csv`, so that it starts in 2 instead of 1, the test still passes. I think that is fine if we are able to verify from the symptom data that ids have to be contained in `two_rate_fns.csv`. Right now, that's not the case, and the ids are not traceable. Should we force these ids to start from 0 or from 1?

confunguido (changes requested):
I think we are missing a test checking that ids from symptom data can be found in `rate_rn_storage`.

confunguido (approved):

PR #108, opened 2025-04-22
Calculate person-level Weibull parameters from isolation guidance Stan parameters

ChiragKumar9:
This PR deals only with the symptom improvement time distributions. We will implement person-level incubation time distributions in the future, but I wanted to get eyes on this as a partial "is this reasonable" flag.


ChiragKumar9:
Pre-commit fails because #95 has not been merged into this branch (since it is a branch off of a branch). We expect cargo to fail at this time because the main model is being run and there are numerical errors in model rate function calculation causing failure (see #88 for details).

eqmooring (approved):
This looks good (made of very minor suggestion). The big chunk of algebra appears to be correct to me, though I have admittedly not cross referenced to the original stan code. I think it'd be good to eventually include a test where distribution of symptom durations in this model is compared to distribution of symptom duration in the output from Stan.

ChiragKumar9:
Thanks, @eqmooring! Yeah this is a great suggestion for a test. Maybe I'll add it to #110 based off work in #88. I will merge this PR into #88 so we can continue work there. I'll get your review on that too, but I wanted the small chunks scrutinized as well.

PR #110, opened 2025-04-24
Read `SymptomData` progressions within a general `read_progressions` framework. Closes #74.

### LINKED ISSUE #74: Add a way to read in empirical symptom progressions from an external file

ChiragKumar9:
This PR proposes a general way to read a progression library (modeled after how we read in a rate function library) and applies the proposed schema to reading in symptom progressions via the `SymptomData` struct.

Most importantly, I elaborate on how we are modeling symptoms. In brief, I argue that we require the incubation period to be fed in as a value for each progression (we do the sampling in a pre-processing step because we require the sampled value to convert the triangle viral load curves from "time since symptom onset" to "time since infection" and then we just feed in the sampled values into Rust), and then we draw symptom improvement times from a Weibull distribution with the parameters. #106 handles ensuring that the right symptom progression is drawn given the rate function that the person was assigned.

Finally, I choose to model the Weibull times as continuous but with an upper bound max -- I want to investigate what is the "right" way to model symptoms (ideally through a calibration) in the future, and I will file this as an issue once we've agreed on what is here.

I slightly refactored the `Progressions` trait to instead be associated with a person property rather than the person property's value, implicitly requiring the user to make that generalization. This allows us to make blanket implementations of `Progressions` on types, usefully for the null implementation.


src/symptom_progression.rs
ChiragKumar9, line 1: @eqmooring This file is really where I want your review.

src/symptom_progression.rs
```diff
@@ -23,28 +28,142 @@ pub enum SymptomValue {
...
+}
+
+impl SymptomData {
+    #[allow(clippy::needless_pass_by_value)]
+    pub fn register(
+        context: &mut Context,
+        parameter_names: Vec<String>,
+        parameters: Vec<f64>,
+    ) -> Result<(), IxaError> {
+        // The first parameter is the symptom category name, the next three are the
```
eqmooring, line 88: should this say
```suggestion
        // The first parameter is the symptom category name, the next one is the
```

src/symptom_progression.rs
eqmooring: One aspect I am trying to understand here is how general/abstract this code is meant to be. My current understanding is that much of this code is specific to an externally determined incubation period followed by an internally drawn right truncated Weibull. Maybe another way to get at this is to ask, if I wanted (hypothetically) a set up where a person had an incubation period drawn from a uniform distribution, followed by an exponentially distributed pauci-symptomatic duration, followed by a Gamma-distributed severely symptomatic phase, it'd be necessary to change quite a bit of this piece of code, right? (Though the changes might be fairly easy, because this could be used as template?) I'm not saying this is bad at all, I just want to understand how general this intended to be right now.

eqmooring:
I think this looks good. There's obviously a lot here. It seems to parallel what we did in the individual duration of isolation analysis. I'd want to also look at how the incubation periods were generated and at how there's alignment between the transmission rate curves and the symptom progressions (I realize those are in different PRs). My overall thinking is that there are lots of ways to model symptoms / symptom progressions. A series of distributions (here a fixed duration followed by a randomly drawn Weibull) makes sense.

eqmooring (approved):

src/infection_propagation_loop.rs
```diff
@@ -104,10 +104,14 @@ mod test {
             initial_infections: 3,
             max_time: 100.0,
             seed,
-            infectiousness_rate_fn: RateFnType::Constant {
+            infectiousness_rate_fn: LibraryType::Constant {
                 rate,
                 duration: 5.0,
             },
+            symptom_progression_library: LibraryType::Constant {
```
eqmooring, line 111: @ChiragKumar9 , I thought I left a comment to this effect but now looking back at this I think I thought about it but didn't. Anyway, I am a little confused about what `rate: 1.0` means in the context of symptom progression. If I am assuming correctly that this `Constant` example is meant to mirror the approach in `infectiousness_rate_fn`, then I am not sure why a rate function is needed.

ChiragKumar9: Yeah, I think I was trying to shove a circular peg into a square hole here. It doesn't really make sense to use the same enum for these two very different use cases -- and having separate enums also lets us get at your desire for being able to specify and switch in different kinds of symptom progressions. I'll change this!

ChiragKumar9: faea910 addresses this -- thanks for pointing it out!

confunguido (changes requested):
Some minor changes. Overall, I think there are some complexities about making this general that we can avoid by making it specific to the symptom data from the file we have already from stan.

src/property_progression_manager.rs
```diff
@@ -81,33 +87,110 @@ impl ContextPropertyProgressionExt for Context {
...
+
+#[derive(Deserialize)]
+struct ProgressionRecord {
+    id: u8,
+    progression_type: ProgressionType,
+    parameter_name: String,
+    parameter_value: f64,
+}
+
+fn add_progressions_from_file(context: &mut Context, file: PathBuf) -> Result<(), IxaError> {
```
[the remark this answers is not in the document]
ChiragKumar9: > Some questions about this function. I was expecting it to deserialize based on a type (in this case SymptomData), but it uses progression record. Is this because the data is long and not wide? If so, how can we make sure that each record has all the necessary fields for symptomdata?

The data are in long format. This is to try to improve generalizability for when we have more than just symptom data progressions. To ensure that each record has all the necessary fields for symptom data, just like we assume the headers have some names when reading data in wide format, we assume that the long format "parameter names" entry have some set names and we query the associated value with a given parameter name, erroring if the parameter name cannot be found: https://github.com/CDCgov/ixa-epi-isolation/blob/8942079af5eb6cb8e9ae8581410f1b4f141e9c3f/src/symptom_progression.rs#L105. Do you think there's a more explicit way to test for all the necessary fields? Here we raise an error if the necessary field isn't found, but that error is only raised when actually assembling the struct. We could try to fail even earlier, though that would maybe require an additional check?

> I am also a bit confused about whether this should be in symptom_progression or in property_progression_manager. I understand that the property progression manager isn't only for symptom progressions but it requires SymptomData to read the progressions from the file.

Yeah, I struggle with this too. For now, I have the general "read file" in the progression manager and the actual "set up symptom data structs" in the symptom progression manager (https://github.com/CDCgov/ixa-epi-isolation/blob/8942079af5eb6cb8e9ae8581410f1b4f141e9c3f/src/property_progression_manager.rs#L147). I agree that figuring out what goes where in this structure is somewhat fuzzy. In an ideal world, I'd like to be able to "contribute" the function that does the struct set up to the progression manager when registering and then the progression manager just automates everything. What do you think about that? If so, I can file it as an issue.

> Finally, one quick change we talked about reverting to was that if we are reading the incubation period as number from this file, then let's also read the time to symptom improvement.

How about I file this as a scientific question of "are we modeling symptom improvement times correctly"?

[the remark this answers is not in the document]
confunguido: 1) I think it's fine as long as it is breaking early enough if the fields can't be found. 
2) Yeah, I think it would make sense to provide the function to the progression manager to read the file. 
3) Yes, file an issue.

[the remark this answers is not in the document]
ChiragKumar9: > 1. I think it's fine as long as it is breaking early enough if the fields can't be found.
Yeah, it breaks as soon as it tries to query those values. This is a good point. I'll audit to see if it causes trouble in the future.
> 2. Yeah, I think it would make sense to provide the function to the progression manager to read the file.
Ok, this seems like something our developer friends can help us with. I filed an issue sketching out the idea so we can focus on landing this in the meantime without that being a blocker. See #129.
> 3. Yes, file an issue.
See #128.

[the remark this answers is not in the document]
ChiragKumar9: > 2. Yeah, I think it would make sense to provide the function to the progression manager to read the file.
Because data containers are private, we have to have the "read from file" function in the same place as the data container to test whether the read from file function actually works. This could be addressed if we had an immutable way to look at values in the data container that was public, and I've commented about this in an issue in core ixa: https://github.com/CDCgov/ixa/issues/311

Issue #107, opened 2025-04-21
Additional checks for initializing settings


erik-rosenstrom:
## Goal
Additional unit tests need to be added to the settings module to ensure that properties of the setting cannot be gotten before the setting is initialized, and that setting types cannot be registered more than once.

## Context
This addresses the [TODO](https://github.com/CDCgov/ixa-epi-isolation/blob/04abbe73b8dbf27ae8ad9f223a1b5872ecb11739/src/settings.rs#L637) in the setting module.

Issue #105, opened 2025-04-21
Add to settings get contact unit test


erik-rosenstrom:
## Goal
An additional unit test needs to be added to verify that when an individual who is not registered to a setting attempts to draw a contact from that setting, None is returned.

## Context
This addresses the [TODO](https://github.com/CDCgov/ixa-epi-isolation/blob/04abbe73b8dbf27ae8ad9f223a1b5872ecb11739/src/settings.rs#L552) in the settings module.

Issue #103, opened 2025-04-21
Registering people to settings validation


erik-rosenstrom:
## Goal
A validation checks needs to be incorporated into the settings module to ensure people can be registered only to settings that are registered.

## Context
This is addressing a [TODO](https://github.com/CDCgov/ixa-epi-isolation/blob/04abbe73b8dbf27ae8ad9f223a1b5872ecb11739/src/settings.rs#L465) in the settings module.

PR #112, opened 2025-04-24
Setting validation 

erik-rosenstrom:
## Summary
This PR addresses three TODO statements documented in issues #103, #105, and #107. The checks largely involve examining the contents of the `SettingsDataContainer` to observe if setting types are registered. 

## What's new
- A new trait extension, `validate_setting_registration` was added to validate setting type registration. 
- The `getting_setting_properties` trait extension is modified to return a `Result<SettingProperties, IxaError>
- Three new tests were added.

## Questions
-  Is there a way to call `self.validate_setting_registration` from within `get_setting_properties`?
- Should the solution to issue #105 throw an `IxaError` or just return none? 

## Updates After Rebasing

- This PR addresses an additional TODO statement documented in issue #114.
- The validation of settings requires that they be registered before individuals can be added to itineraries which have those settings. This meant that three tests in the population loader require settings to be initialized. It also meant that settings has to be initialized prior to the population loader in main.

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -313,18 +313,12 @@ impl ContextSettingExt for Context {
                     "Setting plugin data is none".to_string(),
                 ))?;
 
-        let registered_setting = data_container.setting_types.get(&TypeId::of::<T>());
-
-        if registered_setting.is_none() {
-            return Err(IxaError::from(
+        match data_container.setting_properties.get(&TypeId::of::<T>()) {
+            None => Err(IxaError::from(
                 "Attempting to get properties of unregistered setting type",
-            ));
+            )),
+            Some(properties) => Ok(*properties),
         }
-        let properties = data_container
-            .setting_properties
-            .get(&TypeId::of::<T>())
-            .unwrap();
-        Ok(*properties)
     }
     fn register_setting_type<T: SettingType + 'static>(
         &mut self,
@@ -359,20 +353,22 @@ impl ContextSettingExt for Context {
         for itinerary_entry in &itinerary {
             // TODO: If we are changing a person's itinerary, the person_id should be removed from vector
             // This isn't the same as the concept of being present or not.
-            let registered_setting = container.setting_types.get(&itinerary_entry.setting_type);
-            if registered_setting.is_none() {
-                return Err(IxaError::from(
-                    "Itinerary entry setting type not registered",
-                ));
+            match container.setting_types.get(&itinerary_entry.setting_type) {
+                None => {
+                    return Err(IxaError::from(
+                        "Itinerary entry setting type not registered",
+                    ))
+                }
+                Some(_) => {
+                    container
+                        .members
+                        .entry(itinerary_entry.setting_type)
+                        .or_default()
+                        .entry(itinerary_entry.setting_id)
+                        .or_default()
+                        .push(person_id);
+                }
             }
-            container
-                .members
-                .entry(itinerary_entry.setting_type)
-                .or_default()
-                .entry(itinerary_entry.setting_id)
-                .or_default()
-                .push(person_id);
-            current_setting_ids.push((itinerary_entry.setting_type, itinerary_entry.setting_id));
         }
 
         // Clean up settings that the person is no longer a member of
@@ -757,8 +753,12 @@ mod test {
     fn test_total_infectiousness_multiplier() {
         // Go through all the settings and compute infectiousness multiplier
         let mut context = Context::new();
-        let _ = context.register_setting_type(Home {}, SettingProperties { alpha: 0.1 });
-        let _ = context.register_setting_type(CensusTract {}, SettingProperties { alpha: 0.01 });
+        context
+            .register_setting_type(Home {}, SettingProperties { alpha: 0.1 })
+            .unwrap();
+        context
+            .register_setting_type(CensusTract {}, SettingProperties { alpha: 0.01 })
+            .unwrap();
 
         for s in 0..5 {
             for _ in 0..5 {
@@ -814,8 +814,12 @@ mod test {
         // TODO: What happens if the person isn't registered in the setting?
         let mut context = Context::new();
         context.init_random(42);
-        let _ = context.register_setting_type(Home {}, SettingProperties { alpha: 0.1 });
-        let _ = context.register_setting_type(CensusTract {}, SettingProperties { alpha: 0.01 });
+        context
+            .register_setting_type(Home {}, SettingProperties { alpha: 0.1 })
+            .unwrap();
+        context
+            .register_setting_type(CensusTract {}, SettingProperties { alpha: 0.01 })
+            .unwrap();
 
         let person_a = context.add_person(()).unwrap();
         let person_b = context.add_person(()).unwrap();
@@ -827,10 +831,9 @@ mod test {
         let _ = context.add_itinerary(person_a, itinerary_a);
         let _ = context.add_itinerary(person_b, itinerary_b);
         assert_eq!(
-            person_b,
+            Some(person_b),
             context
-                .get_contact(person_a, SettingId::<Home>::new(0), ())
-                .unwrap()
+                .get_contact::<Home>(person_a, SettingId::<Home>::new(0))
                 .unwrap()
         );
         assert!(context
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -219,6 +219,7 @@ impl ContextSettingInternalExt for Context {
                     "Attempting contact outside of group membership",
                 ));
             }
+            // The setting has one person in it -- this person
             if members.len() == 1 {
                 return Ok(None);
             }
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -212,41 +212,43 @@ impl ContextSettingInternalExt for Context {
         setting_id: usize,
         q: T,
     ) -> Result<Option<PersonId>, IxaError> {
-        let members = self
-            .get_setting_members_internal(setting_type, setting_id)
-            .unwrap();
-        if !members.contains(&person_id) {
-            return Err(IxaError::from(
-                "Attempting contact outside of group membership",
-            ));
-        }
-        if members.len() == 1 {
-            return Ok(None);
-        }
-        let member_iter = members.iter().filter(|&x| *x != person_id);
-
-        let mut contacts = vec![];
-        if q.get_query().is_empty() {
-            // If the query is empty we push members directly to the vector
-            for contact in member_iter {
-                contacts.push(*contact);
+        let members = self.get_setting_members_internal(setting_type, setting_id);
+        if let Some(members) = members {
+            if !members.contains(&person_id) {
+                return Err(IxaError::from(
+                    "Attempting contact outside of group membership",
+                ));
             }
-        } else {
-            // If the query is not empty, we match setting members to the query
-            for contact in member_iter {
-                if self.match_person(*contact, q) {
+            if members.len() == 1 {
+                return Ok(None);
+            }
+            let member_iter = members.iter().filter(|&x| *x != person_id);
+
+            let mut contacts = vec![];
+            if q.get_query().is_empty() {
+                // If the query is empty we push members directly to the vector
+                for contact in member_iter {
                     contacts.push(*contact);
                 }
+            } else {
+                // If the query is not empty, we match setting members to the query
+                for contact in member_iter {
+                    if self.match_person(*contact, q) {
+                        contacts.push(*contact);
+                    }
+                }
             }
-        }
 
-        if contacts.is_empty() {
-            return Ok(None);
-        }
+            if contacts.is_empty() {
+                return Ok(None);
+            }
 
-        Ok(Some(
-            contacts[self.sample_range(SettingsRng, 0..contacts.len())],
-        ))
+            Ok(Some(
+                contacts[self.sample_range(SettingsRng, 0..contacts.len())],
+            ))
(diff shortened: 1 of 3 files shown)


src/settings.rs
```diff
@@ -208,20 +214,30 @@ impl ContextSettingInternalExt for Context {
...
 
 impl ContextSettingExt for Context {
-    fn get_setting_properties<T: SettingType + 'static>(&self) -> SettingProperties {
-        let data_container = self
-            .get_data_container(SettingDataPlugin)
-            .unwrap()
+    fn get_setting_properties<T: SettingType + 'static>(
+        &self,
+    ) -> Result<SettingProperties, IxaError> {
+        let data_container = self.get_data_container(SettingDataPlugin).unwrap();
```
ChiragKumar9, line 220: It seems like we should raise an error if the setting data plugin is None as well -- because that would mean no settings had been registered. So, instead of calling `unwrap` here, I would call `ok_or(Err(IxaError::IxaError("your error message").to_string()))`.

Honestly, I think trying to get setting properties for non registered settings is deserving of a panic rather than an error, but I like having the validation checks so I'm good with it for now. Also, settings will look differently later on, and I think this validation will hopefully become not necessary once we wrap some of the setting data containers in RefCells and do lazy initialization.

src/settings.rs
```diff
@@ -208,20 +214,30 @@ impl ContextSettingInternalExt for Context {
...
+
+        let registered_setting = data_container.setting_types.get(&TypeId::of::<T>());
+
+        if registered_setting.is_none() {
+            return Err(IxaError::from(
+                "Attempting to get properties of unregistered setting type",
```
ChiragKumar9, line 226: If you wanted to be fancy, you could print out what that setting type is!

src/settings.rs
```diff
@@ -208,20 +214,30 @@ impl ContextSettingInternalExt for Context {
...
-        *data_container
+        Ok(*properties)
     }
     fn register_setting_type<T: SettingType + 'static>(
         &mut self,
         setting_type: T,
         setting_props: SettingProperties,
-    ) {
+    ) -> Result<(), IxaError> {
+        self.validate_setting_registration(&setting_type)?;
```
ChiragKumar9, line 240: I would merge the validation into the registration functions. So, when you add the setting, I would do something like follows:
```
match container.setting_types.entry(TypeId::of::<T>()) {
    Entry::Occupied(_) => return Err(IxaError::IxaError(format!("Setting type {:?} already registered", setting_type))),
    Entry::Vacant(entry) => entry.insert(Box::new(setting_type)
}
```

Here's an example straight from the Ixa docs: https://ixa.rs/doc/src/ixa/global_properties.rs.html#229

ChiragKumar9: Since calling `get_data_container_mut` initializes the data container if it is empty, we don't have to do this option handling that we have to do when calling the vanilla `get_data_container` that we do below in the `validate_setting_registration` standalone function.

src/settings.rs
```diff
@@ -311,18 +348,23 @@ impl ContextSettingExt for Context {
         &self,
         person_id: PersonId,
         setting_id: SettingId<T>,
-    ) -> Option<PersonId> {
+    ) -> Result<Option<PersonId>, IxaError> {
         if let Some(members) = self.get_setting_members::<T>(setting_id) {
+            if !members.contains(&person_id) {
```
ChiragKumar9, line 353: I like this validation, I like tests like this, but my question is really just whether an error like this can ever even happen.

src/settings.rs
```diff
@@ -357,19 +399,64 @@ mod test {
...
         assert_almost_eq!(0.001, tract_props.alpha, 0.0);
     }
 
+    #[test]
+    fn test_get_properties_after_registration() {
+        let mut context = Context::new();
+        context
+            .register_setting_type(Home {}, SettingProperties { alpha: 0.1 })
+            .unwrap();
+        let _ = context.get_setting_properties::<Home>().unwrap();
```
ChiragKumar9, line 421:
```suggestion
        context.get_setting_properties::<Home>().unwrap();
```

src/settings.rs
```diff
@@ -357,19 +399,64 @@ mod test {
...
+            ),
+            None => panic!("Expected an error. Instead, validation passed with no errors."),
+        }
+    }
+
     #[test]
     fn test_duplicated_itinerary() {
         let mut context = Context::new();
-        context.register_setting_type(Home {}, SettingProperties { alpha: 1.0 });
+        let _ = context.register_setting_type(Home {}, SettingProperties { alpha: 1.0 });
```
ChiragKumar9, line 459:
```suggestion
        context.register_setting_type(Home {}, SettingProperties { alpha: 1.0 }).unwrap();
```

src/settings.rs
```diff
@@ -390,9 +477,9 @@ mod test {
     }
 
     #[test]
-    fn test_feasible_itinerary() {
+    fn test_feasible_itinerary_ratio() {
         let mut context = Context::new();
-        context.register_setting_type(Home {}, SettingProperties { alpha: 1.0 });
+        let _ = context.register_setting_type(Home {}, SettingProperties { alpha: 1.0 });
```
ChiragKumar9, line 482:
```suggestion
        let _ = context.register_setting_type(Home {}, SettingProperties { alpha: 1.0 });
```

src/settings.rs
```diff
@@ -407,10 +494,31 @@ mod test {
         }
     }
 
+    #[test]
+    fn test_feasible_itinerary_setting() {
+        let mut context = Context::new();
+        let _ = context.register_setting_type(Home {}, SettingProperties { alpha: 1.0 });
```
ChiragKumar9, line 500:
```suggestion
        context.register_setting_type(Home {}, SettingProperties { alpha: 1.0 }).unwrap();
```

src/settings.rs
```diff
@@ -407,10 +494,31 @@ mod test {
...
+            ),
+            None => panic!("Expected an error. Instead, validation passed with no errors."),
+        }
+    }
+
     #[test]
     fn test_add_itinerary() {
         let mut context = Context::new();
-        context.register_setting_type(Home {}, SettingProperties { alpha: 1.0 });
+        let _ = context.register_setting_type(Home {}, SettingProperties { alpha: 1.0 });
```
ChiragKumar9, line 521:
```suggestion
        context.register_setting_type(Home {}, SettingProperties { alpha: 1.0 }).unwrap();
```

src/settings.rs
```diff
@@ -564,16 +672,34 @@ mod test {
         let itinerary_b = vec![ItineraryEntry::new(&SettingId::<Home>::new(0), 1.0)];
         let _ = context.add_itinerary(person_a, itinerary_a);
         let _ = context.add_itinerary(person_b, itinerary_b);
-
         assert_eq!(
             person_b,
             context
                 .get_contact::<Home>(person_a, SettingId::<Home>::new(0))
                 .unwrap()
+                .unwrap()
```
ChiragKumar9, line 680: It looks like there's an extra unwrap here?

ChiragKumar9 (changes requested):
My main comment is about joining the validation in the registration of a setting type, but otherwise it's looking good!

src/settings.rs
erik-rosenstrom: I think someone using the modeling framework, or using the plugins to create their own model, could implement a model where this could happen. They could hard code for everyone to infected people only in the community like is done in the test. This check also seems useful for implementing/debugging more complex uses of settings like where people are switching itineraries frequently. In my opinion, if we can come up with a pre-condition test to catch something we know is wrong it is useful to implement because it is hard to predict how the code will be used down the road by users

src/settings.rs
erik-rosenstrom: I don't think so. `get_contact` returns `Result<Option<PersonId>, IxaError>` which has to be unwrapped twice to extract the person Id. This is not ideal for sure.

src/settings.rs
```diff
@@ -208,32 +211,49 @@ impl ContextSettingInternalExt for Context {
...
+                ))?;
+
+        let registered_setting = data_container.setting_types.get(&TypeId::of::<T>());
+
+        if registered_setting.is_none() {
+            return Err(IxaError::from(
+                "Attempting to get properties of unregistered setting type",
+            ));
+        }
+        let properties = data_container
```
ChiragKumar9, line 230: How about grouping the two `.get(_)` statements into one:

```rust
match data_container.setting_types.get(&TypeId::of::<T>()) {
    None => Err(IxaError::from("Attempting to get properties of unregistered setting type"),
    Some(properties) => Ok(*properties)
```

src/settings.rs
```diff
@@ -244,6 +264,12 @@ impl ContextSettingExt for Context {
         for itinerary_entry in &itinerary {
             // TODO: If we are changing a person's itinerary, the person_id should be removed from vector
             // This isn't the same as the concept of being present or not.
+            let registered_setting = container.setting_types.get(&itinerary_entry.setting_type);
+            if registered_setting.is_none() {
```
ChiragKumar9, line 268: Same proposal for condensing the result of a `.get(_)` as above.

src/settings.rs
ChiragKumar9: This is fair!

src/settings.rs
```diff
@@ -495,8 +612,8 @@ mod test {
     fn test_total_infectiousness_multiplier() {
         // Go through all the settings and compute infectiousness multiplier
         let mut context = Context::new();
-        context.register_setting_type(Home {}, SettingProperties { alpha: 0.1 });
-        context.register_setting_type(CensusTract {}, SettingProperties { alpha: 0.01 });
+        let _ = context.register_setting_type(Home {}, SettingProperties { alpha: 0.1 });
```
ChiragKumar9, line 615: unwrap here too?

src/settings.rs
ChiragKumar9: Ah, I would check that the result of one unwrap equals `Some(person_b)` to improve readability

src/settings.rs
```diff
@@ -592,8 +727,9 @@ mod test {
         for seed in 0..100 {
             let mut context = Context::new();
             context.init_random(seed);
-            context.register_setting_type(Home {}, SettingProperties { alpha: 0.1 });
-            context.register_setting_type(CensusTract {}, SettingProperties { alpha: 0.01 });
+            let _ = context.register_setting_type(Home {}, SettingProperties { alpha: 0.1 });
```
ChiragKumar9, line 730: unwrap here if you'd like

ChiragKumar9 (approved):
I think this looks good! Small clean up proposed, but I don't need to see it again for review. Thanks for working through this!

src/settings.rs
```diff
@@ -238,35 +241,42 @@ impl ContextSettingInternalExt for Context {
...
-                    for contact in member_iter {
-                        if self.match_person(*contact, q) {
-                            contacts.push(*contact);
-                        }
-                    }
-                }
+    ) -> Result<Option<PersonId>, IxaError> {
+        let members = self
+            .get_setting_members_internal(setting_type, setting_id)
+            .unwrap();
```
ChiragKumar9, line 247: I don't think this should be an unwrap. If it's an unwrap, when `get_setting_members_internal` returns None, this code panics rather than returning `Ok(None)`. I think you can keep the `.and_then` structure and just add a check that confirms that `members` contains the person_id and if it doesn't return an error.

src/settings.rs
```diff
@@ -657,9 +804,8 @@ mod test {
 
     #[test]
     fn test_setting_multiplier() {
-        // TODO: if setting not registered, shouldn't be able to register people to setting
         let mut context = Context::new();
-        context.register_setting_type(Home {}, SettingProperties { alpha: 0.1 });
+        let _ = context.register_setting_type(Home {}, SettingProperties { alpha: 0.1 });
```
ChiragKumar9, line 808: In all of these `let _ = ` places, I would just `.unwrap()` instead

ChiragKumar9 (changes requested):
My main change is in the `get_contact_internal` method.

src/settings.rs
ChiragKumar9: But it's also really up to you truth be told

src/settings.rs
erik-rosenstrom: rather than worrying about the and_then implementation should we throw an error if the let members = self.get_setting_members_internal(setting_type, setting_id); is none when calling the get get_contact_internal function. It seems like something would be wrong if thats the case. It is similar to if the person_Id is not in the membership list.

src/settings.rs
```diff
@@ -208,35 +211,44 @@ impl ContextSettingInternalExt for Context {
...
-                let member_iter = members.iter().filter(|&x| *x != person_id);
+    ) -> Result<Option<PersonId>, IxaError> {
+        let members = self.get_setting_members_internal(setting_type, setting_id);
+        if let Some(members) = members {
+            if !members.contains(&person_id) {
+                return Err(IxaError::from(
+                    "Attempting contact outside of group membership",
+                ));
+            }
+            if members.len() == 1 {
```
ChiragKumar9, line 222:
```suggestion
            // The setting has one person in it -- this person
            if members.len() == 1 {
```

ChiragKumar9 (approved):
Thanks for working through all this! What's been scoped out in this PR looks good to me; I would consider (and I know this code predates your PR) but merging `validate_itinerary`'s public method into your internal private validation methods as a follow-up PR if you'd like.

PR #120, opened 2025-04-30
Intervention module - transmission modifiers

KOVALW:
We require methods to modify the relative transmission potential of given infection attempts and the overall rate of infection attempt creation


src/settings.rs
ChiragKumar9, line 1: Changes to this module would be removed if the infectiousness modifier were not incorporated into settings.

ChiragKumar9: More to the point, here we're calling it the "intrinsic" infectiousness modifier when it's really the "total" infectiousness modifier, so I am starting to feel even more strongly that we're mixing apples and oranges.

src/infection_propagation_loop.rs
```diff
@@ -17,11 +19,15 @@ fn schedule_next_forecasted_infection(context: &mut Context, person: PersonId) {
     }) = get_forecast(context, person)
     {
         context.add_plan(next_time, move |context| {
-            // TODO<ryl8@cc.gov>: We will choose a setting here
             if evaluate_forecast(context, person, forecasted_total_infectiousness) {
-                if let Some(next_contact) = infection_attempt(context, person) {
-                    trace!("Person {person}: Forecast accepted, infecting {next_contact}");
-                    context.infect_person(next_contact, Some(person));
+                if let Some(next_contact) = context
+                    .draw_contact_from_transmitter_itinerary(person, (Alive, true))
```
ChiragKumar9, line 24: I'm fine with this change, but I'm somewhat curious about why? It feels more natural to me to have `infection_attempt` also include the contact drawing especially if contact drawing will be split into two parts -- setting choice and then contact selection.

src/infection_propagation_loop.rs
```diff
@@ -393,4 +428,181 @@ mod test {
...
+        }
+
+        // Degrees of freedom = number of bins - 1
+        let degrees_of_freedom = observed_counts.len() - 1;
+        let chi_square_critical = statrs::distribution::ChiSquared::new(degrees_of_freedom as f64)
+            .unwrap()
+            .inverse_cdf(0.95);
+
+        assert!(
+            chi_square_stat < chi_square_critical,
```
ChiragKumar9, line 604: Our convention is to be a little bit less hypothesis test-y about this. Just say that `chi_square_stat` is less than some user-set critical value and do a sniff test that that value decreases as you run more iterations.

src/infection_propagation_loop.rs
```diff
@@ -393,4 +428,181 @@ mod test {
...
+        set_homogeneous_mixing_itinerary(&mut context, infectious_person).unwrap();
+        for _ in 0..community_size - 1 {
+            let cohabitant = context.add_person(()).unwrap();
+            set_homogeneous_mixing_itinerary(&mut context, cohabitant).unwrap();
+        }
+
+        // Listen for infection events
+        // We do not schedule the next forecasted infection for the secondary cases, as we want to see how many
+        // secondary cases are created by only the initial infectious person before they recover
+        let secondary_cases = Rc::new(RefCell::new(vec![]));
```
ChiragKumar9, line 520: Can we name this vector more clearly as `secondary_infection_times` or something?

src/infection_propagation_loop.rs
```diff
@@ -393,4 +428,181 @@ mod test {
...
+        let returned_vec = secondary_cases.borrow().clone();
+        returned_vec
+    }
+
+    #[test]
+    #[allow(clippy::cast_lossless, clippy::cast_precision_loss)]
+    fn test_community_masking_one_infector() {
+        let rate = 2.0;
+        let alpha = 0.1;
+        let duration = 4.0;
```
ChiragKumar9, line 549: I don't understand why you need a duration of 4 days here rather than 5 days. In general, I am a fan of trying to change as few things as possible when it comes to the context set up.

src/infection_propagation_loop.rs
```diff
@@ -393,4 +428,181 @@ mod test {
...
+        let expected_cases = (1.0 - masking_proportion)
+            * rate
+            * duration
+            * ((community_size - 1) as f64).powf(alpha);
+
+        // Perform a goodness-of-fit test to check if the case count sizes follow a Poisson distribution
+        let poisson = Poisson::new(expected_cases).unwrap();
+        let mut observed_counts = vec![0; community_size];
+        for &count in &cases_count_sizes {
+            if count < community_size {
```
ChiragKumar9, line 580: I think I understand what's going on with this test. I want to take a look at it again and try to better understand what's going on in the revision. I'm also not sure this test introduces a new test. We already test infection times -- I don't know what else is being tested here besides -- in essence -- showing that the infection rates change with the presence of a modifier.

src/interventions/transmission_modifier_manager.rs
```diff
@@ -0,0 +1,394 @@
...
+define_data_plugin!(
+    TransmissionModifierPlugin,
+    TransmissionModifierContainer,
+    TransmissionModifierContainer {
+        transmission_modifier_map: HashMap::new(),
+        modifier_aggregator: HashMap::new(),
+    }
+);
+
+pub trait ContextTransmissionModifierExt {
```
ChiragKumar9, line 51: Can you also file that we need to make a generic `register_modifier_fn` method too?

src/interventions/transmission_modifier_manager.rs
```diff
@@ -0,0 +1,394 @@
...
+    /// Register a transmission modifier for a specific infection status and person property.
+    /// Modifier key must have specified lifetime to outlive the Box'd `TrasnmissionModifierFn`.
+    /// Modifier key is taken as a slice to avoid new object creation through Vec
+    fn register_transmission_modifier<T: PersonProperty + 'static>(
```
ChiragKumar9, line 55: To incorporate setting-specificity of intervention efficacy, I would have this method take the setting type in the future as well (well, maybe make a new method that also takes the setting type in the modifier map) and calculate setting-specific efficacy in the hook rather than deferring that calculation into the `get_multiplier` method as you are templating now. But, more broadly, I would make sure that we are separating total infectiousness modifiers and intrinsic infectiousness modifiers (and then separately itinerary modifiers).

src/interventions/transmission_modifier_manager.rs
```diff
@@ -0,0 +1,394 @@
...
+        &self,
+        infection_status: InfectionStatusValue,
+        modifiers: &Vec<(TypeId, f64)>,
+    ) -> f64 {
+        self.modifier_aggregator
+            .get(&infection_status)
+            .unwrap_or(&Self::default_aggregator())(modifiers)
+    }
+
+    fn default_aggregator() -> Box<TransmissionAggregatorFn> {
```
RobertJacobsonCDC, line 29: Since you just turn around and call the default aggregator, why not avoid the heap allocation? In other words, do this:

```rust
impl TransmissionModifierContainer {
    fn run_aggregator(
        &self,
        infection_status: InfectionStatusValue,
        modifiers: &Vec<(TypeId, f64)>,
    ) -> f64 {
        self
            .modifier_aggregator
            .get(&infection_status)
            .map(|agg| agg(modifiers))
            .unwrap_or_else(|| default_aggregator(modifiers))
    }
}

// a regular function, no heap involved
fn default_aggregator(modifiers: &[(TypeId, f64)]) -> f64 {
    modifiers.iter().map(|&(_, eff)| eff).product()
}
```

src/interventions/transmission_modifier_manager.rs
```diff
@@ -0,0 +1,394 @@
...
+    /// and its corresponding modifier value.
+    /// The default aggregator multiplies all the modifier values together independently.
+    #[allow(dead_code)]
+    fn register_transmission_aggregator(
+        &mut self,
+        infection_status: InfectionStatusValue,
+        agg_function: Box<TransmissionAggregatorFn>,
+    );
+
+    /// Get the relative intrinsic transmission (infectiousness or susceptiblity) for a person based on their
```
RobertJacobsonCDC, line 73: susceptiblity -> susceptibility

src/interventions/transmission_modifier_manager.rs
```diff
@@ -0,0 +1,394 @@
...
+}
+
+impl ContextTransmissionModifierExt for Context {
+    fn register_transmission_modifier<T: PersonProperty + 'static>(
+        &mut self,
+        infection_status: InfectionStatusValue,
+        person_property: T,
+        modifier_key: &'static [(T::Value, f64)],
+    ) {
+        let transmission_modifier_map: HashMap<TypeId, Box<TransmissionModifierFn>> =
```
RobertJacobsonCDC, line 85: Instead of creating this every single time, put it into a closure and only call the closure when the entry isn't found. I think the method you want is `or_insert_with()`:

```rust
fn register_transmission_modifier<T: PersonProperty + 'static>(
        &mut self,
        infection_status: InfectionStatusValue,
        person_property: T,
        modifier_key: &'static [(T::Value, f64)],
    ) {
        self.get_data_container_mut(TransmissionModifierPlugin)
            .transmission_modifier_map
            .entry(infection_status)
            .or_insert_with(|| {
                HashMap::from([(
                    TypeId::of::<T>(),
                    Box::new(move |context: &Context, person_id: PersonId| -> f64 {
                        let property_val = context.get_person_property(person_id, person_property);
                        for item in modifier_key {
                            if property_val == item.0 {
                                return item.1;
                            }
                        }
                        // Return a default 1.0 (no relative change if unregistered)
                        1.0
                    })
                )])
            });
    }
```

src/interventions/transmission_modifier_manager.rs
```diff
@@ -0,0 +1,394 @@
...
+            .transmission_modifier_map
+            .entry(infection_status)
+            .or_default()
+            .extend(transmission_modifier_map);
+    }
+
+    fn register_transmission_aggregator(
+        &mut self,
+        infection_status: InfectionStatusValue,
+        agg_function: Box<TransmissionAggregatorFn>,
```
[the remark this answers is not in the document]
RobertJacobsonCDC: The problem is you need boxing if you are going to take a `dyn trait`, which is what the `TransmissionAggregatorFn` is. In this case I think you need it.

RobertJacobsonCDC (changes requested):
I like the design, personally. I only have superficial comments on the implementation.

Here's another general comment that doesn't really fit anywhere else:

As a matter of design, I think it's better to put the private implementation of the `transmission_modifier_container` in a method on `TransmissionModifierContainer` instead of in the body of `ContextTransmissionModifierExt`, reserving `ContextTransmissionModifierExt` just for defining the public API. In other words, the methods on `ContextTransmissionModifierExt` almost always just delegates to some method on `TransmissionModifierContainer`. For example, working with references owned by the container is a lot easier, because you can't return a reference from a function without something like `Ref`/`RefMut`, which can be complicated.

src/interventions/transmission_modifier_manager.rs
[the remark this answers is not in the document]
ChiragKumar9: You can make the function `register_transmission_aggregator` be generic and take a type that implements the transmission aggregator function. Then, you box that function in the manager, changing the burden of boxing from the user to the manager. I do this in the natural history manager! https://github.com/CDCgov/ixa-epi-isolation/blob/987f9b6bb2dd0f7f4bdc9634b28f31b1555e76a8/src/natural_history_parameter_manager.rs#L72

Here's what it would look like in this example:

```rust
fn register_transmission_aggregator<A>(
    &mut self,
    infection_status: InfectionStatusValue,
    agg_function: A)
where A: Fn(Vec<(TypeId, f64)>) -> f64 {

    // type annotation not necessary -- just for explicitness
    let boxed_agg_function: Box<TransmissionAggregatorFn> = Box::new(agg_function);

}
```

src/interventions/transmission_modifier_manager.rs
```diff
@@ -0,0 +1,394 @@
...
+            registered_modifiers.push((*t, f(self, person_id)));
+        }
+
+        transmission_modifier_plugin.run_aggregator(infection_status, &registered_modifiers)
+    }
+}
+
+// Initialize the transmission modifier plugin with guaranteed values
+pub fn init(context: &mut Context) {
+    context.register_transmission_modifier(
```
confunguido, line 142: I don't get the purpose of this registration. Can you elaborate, Will?

src/interventions/transmission_modifier_manager.rs
```diff
@@ -0,0 +1,394 @@
...
+                (MandatoryIntervention::Partial, SUSCEPTIBLE_PARTIAL),
+                (MandatoryIntervention::Full, 0.0),
+            ],
+        );
+        context.register_transmission_modifier(
+            InfectionStatusValue::Infectious,
+            MandatoryInterventionStatus,
+            &[
+                (MandatoryIntervention::Partial, INFECTIOUS_PARTIAL),
+                (MandatoryIntervention::Full, 0.0),
```
confunguido, line 256: I'm guessing 0.0 here means 100% protective. Is that right?

src/interventions/transmission_modifier_manager.rs
KOVALW: We need some way to provide the user with access to a transmission modifier lookup table without too many nested layers being called all at once from the data container. This could be done with a flat `HashMap`, with `T::Value`s being unique across `PersonProperty`s, but it makes more sense to declare and access the properties one at a time. 

I honestly think that this first implementation makes little sense because we effectively only offer the user a more abstract way to make a nested `HashMap`, so instead in a change I'm about to push, I've split this function into two pieces (a generalized function that registers actual transmission modifier functions, and a specific helper function that stores modifier values by registering a lookup table modifier function). The split helps to make it more clear why the registration of the function is necessary, and provides the user with a convenient place to use the most common modifier assignment method (a lookup table)

src/interventions/transmission_modifier_manager.rs
KOVALW: That's right

src/interventions/transmission_modifier_manager.rs
KOVALW: I've opted for `or_default().insert(TypeId, fn)` because `or_insert_with` wouldn't register the modifiers if the `Entry` was `Occupied`

src/interventions/transmission_modifier_manager.rs
KOVALW: Just went ahead and provided it here

src/infection_propagation_loop.rs
KOVALW: This is right, as I subtly mention in the comments, this is not really distinct form the one person function. The times need to be verified to not overlap with when the infectious individual is wearing a mask if it were to demonstrate something new

src/interventions/transmission_modifier_manager.rs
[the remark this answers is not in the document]
KOVALW: I've gone with this box-inside-function appraoch for both the `TransmissionModifierFn` and for the `TransmissionAggregatorFn` use cases

src/interventions/transmission_modifier_manager.rs
```diff
@@ -0,0 +1,394 @@
...
+        transmission_modifier_container
+            .modifier_aggregator
+            .insert(infection_status, agg_function);
+    }
+
+    fn get_relative_intrinsic_transmission_person(&self, person_id: PersonId) -> f64 {
+        let infection_status = self.get_person_property(person_id, InfectionStatus);
+
+        let transmission_modifier_plugin =
+            self.get_data_container(TransmissionModifierPlugin).unwrap();
```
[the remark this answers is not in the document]
KOVALW: I've added an `if let Some` but will keep the `init` for now. `Alive==false` should always be rejected; even if we hard code this into the infection propagation loop I think it makes sense to keep.

src/interventions/transmission_modifier_manager.rs
```diff
@@ -0,0 +1,394 @@
...
+            context.get_relative_intrinsic_transmission_person(person_id),
+            INFECTIOUS_PARTIAL
+        );
+
+        context.set_person_property(
+            person_id,
+            InfectiousnessReductionStatus,
+            Some(InfectiousnessReduction::Partial),
+        );
+        assert_eq!(
```
confunguido: So, here, you are registering two different transmission modifiers in infectiousness and the aggregator should be the multiplication of both effects because of the default aggregator?

KOVALW: That's right. I've added a new test on the aggregator that makes this result more clear. This test now serves to demonstrate that someone's infectiousness modifier can change when their person properties are updated

src/infectiousness_manager.rs
```diff
@@ -53,29 +53,31 @@ define_derived_property!(
 /// for a person, given factors related to their environment, such as the number of people
 /// they come in contact with or how close they are.
 /// This is used to scale the intrinsic infectiousness function of that person.
+/// All modifiers of the infector's intrinsic infecitousness are calculated and passed
+/// to the `calculate_multiplier` Setting method
 pub fn calc_total_infectiousness_multiplier(context: &Context, person_id: PersonId) -> f64 {
-    // TODO: calculate current vs total infectiousness. This depends on interventions.
-    context.calculate_total_infectiousness_multiplier_for_person(person_id)
+    let modifier = context.get_relative_intrinsic_transmission_person(person_id);
```
confunguido, line 59: I'm not sure that this is the best place to incorporate this modifier. It makes infectiousness manager dependent on having interventions in the model. Have we considered adding this to wherever the total infectiousness multiplier is calculated?

src/interventions/transmission_modifier_manager.rs
```diff
@@ -0,0 +1,394 @@
...
+            let target = context
+                .add_person((MandatoryInterventionStatus, MandatoryIntervention::NoEffect))
+                .unwrap();
+
+            set_homogeneous_mixing_itinerary(&mut context, infector).unwrap();
+            set_homogeneous_mixing_itinerary(&mut context, target).unwrap();
+
+            context.infect_person(infector, None);
+            let forecast = get_forecast(&context, infector).unwrap();
+            if evaluate_forecast(
```
confunguido, line 516: It took me a very long time to understand what happened here since modifier is also affecting `evaluate_forecast`. However, that's not implemented here, so why should you test that here? Another question is, when you calculate the forecast, but no time has passed, what happens inside `evaluate_forecast`?

confunguido (changes requested):
This is a partial review. Take a look at the comments and I will re-review once addressed.

src/infectiousness_manager.rs
KOVALW: I think the rejection sampling for handling Susceptible contacts is going to stay in this manager, so either way there would be dependency here on the intervention manager and we provide 1.0 default values everywhere. 

I've added a new function to the trait extension in settings, where the original calculate multiplier function is, to do this same functionality there instead, though

src/interventions/transmission_modifier_manager.rs
KOVALW: `evaluate_forecast` uses `get_current_total_infectiousness` while `get_forecast` uses `get_max_total_infectiousness`, so no time needs to pass if the properties aren't changing. I'll move this test to the infectiousness manager though because you're right about what it tests

src/interventions/transmission_modifier_manager.rs
```diff
@@ -0,0 +1,394 @@
+use ixa::{define_data_plugin, Context, ContextPeopleExt, PersonId, PersonProperty};
+use std::{any::TypeId, collections::HashMap};
+
+use crate::{
+    infectiousness_manager::{InfectionStatus, InfectionStatusValue},
+    population_loader::Alive,
+};
+
+type TransmissionModifierFn = dyn Fn(&Context, PersonId) -> f64;
+type TransmissionAggregatorFn = dyn Fn(&Vec<(TypeId, f64)>) -> f64;
```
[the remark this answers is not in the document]
KOVALW: Created a draft of this Issue #143

src/infection_propagation_loop.rs
KOVALW: Infection attempt to me signifies attempting to infect a particular contact, which is then assessed. Splitting the draw contact into setting and contact is an even better reason that we want to distinguish between "attempt to infect anyone in your shared settings" and "attempt to infect this person"

src/infection_propagation_loop.rs
```diff
@@ -393,4 +430,205 @@ mod test {
...
+            for &count in &cases_per_masking_period {
+                if count < 10 {
+                    case_count_distribution[count] += 1;
+                }
+            }
+        }
+
+        // Finally, we check that the distribution of case counts is approximately Poisson distributed using the pmf
+        let poisson_dist = Poisson::new(expected_cases / mask_changes).unwrap();
+        for (i, &counts) in case_count_distribution.iter().enumerate() {
```
ChiragKumar9, line 624: Curious why you are not using the KS formulation to test here. A deviation of 0.05 at each probability could actually be a very different distribution, right?

src/infection_propagation_loop.rs
```diff
@@ -393,4 +430,205 @@ mod test {
...
+        // Set up the intervention of facemask and register a perfect facemask modifier
+        context
+            .store_transmission_modifier_values(
+                InfectionStatusValue::Infectious,
+                MaskingStatus,
+                &[(Masking::Wearing, 0.0)],
+            )
+            .unwrap();
+
+        // Switch back and forth between masking or not masking every `masking_duration`, if mask changes is 10.0,
```
ChiragKumar9, line 479:
```suggestion
        // Switch back and forth between masking or not masking every `masking_duration`. If `mask_changes` is 10.0,
```

src/infection_propagation_loop.rs
```diff
@@ -393,4 +430,205 @@ mod test {
...
+                    let period_id =
+                        (context.get_current_time() * mask_changes / duration).trunc() as usize;
+                    secondary_infection_times_clone
+                        .borrow_mut()
+                        .push((period_id, context.get_current_time()));
+
+                    let InfectionDataValue::Infectious { infected_by, .. } =
+                        context.get_person_property(event.person_id, InfectionData)
+                    else {
+                        panic!("Unexpected InfectionDataValue variant encountered");
```
ChiragKumar9, line 542:
```suggestion
                        panic!("{} is not infected.", event.person_id);
```

src/infection_propagation_loop.rs
```diff
@@ -393,4 +430,205 @@ mod test {
...
+                        event.person_id,
+                        InfectionData,
+                        InfectionDataValue::Susceptible,
+                    );
+                }
+            },
+        );
+
+        // Shut down manually to avoid infinite plan loop from masking application
+        context.add_plan(duration, move |context| {
```
ChiragKumar9, line 558: Should this be `add_plan_with_phase` technically? And, shutdown can be written as `context.add_plan(duration, ixa::context::Shutdown)`.

src/infection_propagation_loop.rs
```diff
@@ -393,4 +430,205 @@ mod test {
...
+            context.shutdown();
+        });
+
+        context.execute();
+
+        // The distirbution of infection times should be derived and is returned here for convenience
```
ChiragKumar9, line 564:
```suggestion
        // The distribution of infection times should be derived and is returned here for convenience
```

ChiragKumar9: What do you mean by "should be derived"?

ChiragKumar9 (approved):
I think this is good. But, there's a lot going on here. I plan on going back to take another look at this tests in the infection propagation loop -- in my dream world, your tests are condensed with some of mine about the raw timing of infection events, but I understand if that's not possible at this time.

docs/transmission-modifiers.md
```diff
@@ -0,0 +1,54 @@
...
+Before infecting a contact, an infectious individual must have a successfully scheduled infection
+forecast and the contact must be identified as a successful infection attempt. Factors that change
+the probability of accepting or rejecting either a forecast or scheduled infection attempt can 
+change over time, such as when an infectious individual changes their use of a facemask over the 
+course of infection. Modifiers on the total infectiousness of the infector and susceptibility of 
+the contact therefore need to be identified at runtime.
+
+To include transmission modifiers, the `transmission_modifier_manager.rs` provides the capacity to 
+calculate a single `float` value, less than or equal to one, that represents the probability of 
+accepting a forecasst when calculated for an infectious individual and represents the probability
```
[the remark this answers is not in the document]
KOVALW: These are good points. I wanted to include some amount of new text in the docs for this PR to contextualize things a bit while writing the code rather than playing 100% catch up down the line

src/interventions/transmission_modifier_manager.rs
```diff
@@ -0,0 +1,653 @@
...
+            transmission_modifier_plugin.run_aggregator(infection_status, &registered_modifiers)
+        } else {
+            // If the plugin is not initialized, return 1.0
+            1.0
+        }
+    }
+}
+
+// Initialize the transmission modifier plugin with guaranteed values
+pub fn init(context: &mut Context) -> Result<(), IxaError> {
```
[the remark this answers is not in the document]
KOVALW: I'll cut it, users should rely on the docs, in the future, to see examples and you're correct that we don't init these maps for `InfectionStatus`

src/interventions/transmission_modifier_manager.rs
```diff
@@ -0,0 +1,653 @@
...
+    /// Get the relative intrinsic transmission (infectiousness or susceptibility) for a person based on their
+    /// infection status and current properties based on registered modifiers.
+    fn get_modified_relative_total_transmission_person(&self, person_id: PersonId) -> f64;
+}
+
+impl ContextTransmissionModifierExt for Context {
+    fn register_transmission_modifier_fn<T: PersonProperty + 'static + std::fmt::Debug, F>(
+        &mut self,
+        infection_status: InfectionStatusValue,
+        person_property: T,
```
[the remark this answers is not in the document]
ChiragKumar9: This and other problems filed as #145.

src/infectiousness_manager.rs
```diff
@@ -405,4 +413,89 @@ mod test {
...
+                &[
+                    (MandatoryIntervention::NoEffect, 1.0),
+                    (MandatoryIntervention::Partial, relative_effect),
+                    (MandatoryIntervention::Full, 0.0),
+                ],
+            )
+            .unwrap();
+
+        for _ in 0..n {
+            // Will return None if forecast is greater than infection duration, running 10_000 times encounters unwrap on None
```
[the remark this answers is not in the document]
KOVALW: With 10,000 replicates of this experiment I actually got a forecast to be None, meaning the exponential draw was longer than the infecitous period duration, and when unwrapping the forecast it panicked. This also means that our test for `get_forecast` would stochastically fail, maybe 0.01% of the time.

src/infectiousness_manager.rs
```diff
@@ -405,4 +413,89 @@ mod test {
...
+            let f = get_forecast(&context, p1).expect("Forecast should be returned");
```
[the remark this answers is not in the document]
KOVALW: The previous test calls `infection_attempt`, which assesses suitability of a contact for infection, but what we're testing here is the forecast rejection rate, which is a slimmed down version of `schedule_next_forecasted_infection`

src/infection_propagation_loop.rs
KOVALW: It shouldn't matter because no events occur at or after duration that we care about

src/infection_propagation_loop.rs
KOVALW: This was a note to self that escaped my memory - deleted.

src/settings.rs
```diff
@@ -440,6 +451,11 @@ impl ContextSettingExt for Context {
         collector
     }
 
+    fn calculate_effective_infectiousness_multiplier_for_person(&self, person_id: PersonId) -> f64 {
```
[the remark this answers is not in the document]
KOVALW: moved back to be modifier * total infectiousness in the infectiousness manager and deleted this method in the trait

docs/transmission-modifiers.md
```diff
@@ -0,0 +1,54 @@
+## Modifiers on total transmission
+Before infecting a contact, an infectious individual must have a successfully scheduled infection
+forecast and the contact must be identified as a successful infection attempt. Factors that change
+the probability of accepting or rejecting either a forecast or scheduled infection attempt can
+change over time, such as when an infectious individual changes their use of a facemask over the
+course of infection. Modifiers on the total infectiousness of the infector and susceptibility of
+the contact therefore need to be identified at runtime.
```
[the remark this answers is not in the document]
KOVALW: Not quite what I'm saying modifiers have to all be observable to the aggregator at run time, and will otherwise be excluded, so we need a dynamic way that doesn't depend on the user being explicit about which interventions are active in more than one place, or even at compile time given this will likely be an input file

src/interventions/transmission_modifier_manager.rs
```diff
@@ -0,0 +1,643 @@
...
+use std::{any::TypeId, collections::HashMap};
+
+use crate::infectiousness_manager::{InfectionStatus, InfectionStatusValue};
+
+type TransmissionModifierFn = dyn Fn(&Context, PersonId) -> f64;
+type TransmissionAggregatorFn = dyn Fn(&Vec<(TypeId, f64)>) -> f64;
+
+struct TransmissionModifierContainer {
+    transmission_modifier_map:
+        HashMap<InfectionStatusValue, HashMap<TypeId, Box<TransmissionModifierFn>>>,
```
RobertJacobsonCDC, line 13: It seems to me `transmission_modifier_map` wants to be a `HashMap<(InfectionStatusValue, TypeId), Box<TransmissionModifierFn>>`, or, for multiple modifiers per property, `HashMap<(InfectionStatusValue, TypeId), Vec<Box<TransmissionModifierFn>>>`. This would simplify some of your code.

RobertJacobsonCDC (changes requested):

RobertJacobsonCDC (approved):

src/infectiousness_manager.rs
```diff
@@ -53,29 +53,34 @@ define_derived_property!(
...
-        .unwrap()?;
-    match context.get_person_property(next_contact, InfectionStatus) {
-        InfectionStatusValue::Susceptible => Some(next_contact),
-        _ => None,
+// Infection attempt function to perform rejection sampling on particular contact given modifiers
+pub fn infection_attempt(context: &Context, contact: PersonId) -> bool {
+    match context.get_person_property(contact, InfectionStatus) {
+        InfectionStatusValue::Susceptible => {
+            let relative_transmission_success =
+                context.get_modified_relative_total_transmission_person(contact);
```
confunguido, line 80: If I understand correctly, this intervention here will apply on the susceptibility of `contact` . That's because `get_modified_relative_total_transmission_person` looks for the intervention on the current value of the person property. Am I getting that right?

KOVALW: Right - the `get_modified_relative_total_transmission_person` knows to look for `Susceptible` as the `InfectionStatus` and will search the modifier functions mapped to the susceptibility only

src/interventions/transmission_modifier_manager.rs
```diff
@@ -0,0 +1,394 @@
...
+    TransmissionModifierContainer {
+        transmission_modifier_map: HashMap::new(),
+        modifier_aggregator: HashMap::new(),
+    }
+);
+
+pub trait ContextTransmissionModifierExt {
+    /// Register a transmission modifier for a specific infection status and person property.
+    /// Modifier key must have specified lifetime to outlive the Box'd `TrasnmissionModifierFn`.
+    /// Modifier key is taken as a slice to avoid new object creation through Vec
```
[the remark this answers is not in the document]
confunguido: Agree with this. I think we need a better name than `modifier`. A modifier of value 1.0 sounds like it modifies 100% of transmission.

src/infection_propagation_loop.rs
```diff
@@ -267,6 +272,18 @@ mod test {
         );
     }
 
+    #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, Hash)]
+    pub enum InfectiousnessReduction {
+        Partial,
+    }
+    define_person_property_with_default!(
```
confunguido, line 279: Same comment about variable names. I am getting confused. Does an infectiousnessReduction of 1.0 imply no reduction or 100% reduction? Please make those variable names more clear. If a modifier value of 1.0 is 0% reduction, it should have a different name.

src/infection_propagation_loop.rs
```diff
@@ -393,4 +429,203 @@ mod test {
...
+    }
+
+    #[test]
+    #[allow(
+        clippy::cast_lossless,
+        clippy::cast_precision_loss,
+        clippy::cast_sign_loss,
+        clippy::cast_possible_truncation
+    )]
+    fn test_community_masking_one_infector() {
```
confunguido, line 573: This is a nice test. What would happen if an infectious person wears a facemask for 50% of their infectious period? How does that affect the forecast and the infection attempt?

confunguido (changes requested):
I have a few remaining comments but overall looks good.

src/infection_propagation_loop.rs
KOVALW: Can you provide some more detail on what you're asking here? The test currently runs with 40% of the infectious period being masked. As I understand it, your 50% question sounds more like a scenarios/scientific question rather than a code functionality, so I might be missing something.

confunguido (approved):

PR #122, opened 2025-04-30
Make empirical rates not extrapolate rate below the minimum provided time. Closes #121.

### LINKED ISSUE #121: Make infectiousness pre `times[0]` be 0 in empirical rate
@eqmooring rightly pointed out that it is confusing to a user to assume that infectiousness rate is 0 after the last time in the empirical rate function time vector but that infectiousness before the first time in the vector may be interpolated from values in the vector. Instead, we propose making the infectiousness rate function be 0 for times before the first value in the times vector.

ChiragKumar9:
See #121 for details/description/problem.

### DIFF
--- a/src/rate_fns/empirical_rate.rs
+++ b/src/rate_fns/empirical_rate.rs
@@ -275,6 +275,8 @@ fn get_lower_index(xs: &[f64], xp: f64) -> (usize, usize) {
 
 #[cfg(test)]
 mod test {
+    use core::f64;
+
     use ixa::IxaError;
     use statrs::assert_almost_eq;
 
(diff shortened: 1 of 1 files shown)


eqmooring (approved):
I'm certainly not a Rust expert, but the changes make sense to me. Moreover, I rebased my branch on this one, re-ran the SEIR-like simulation (without the little hack of having a zero rate slightly before infectiousness starts), and--by visual inspection--the results look good. (I do see that you had documented infectiousness before the first value in the time series, but I missed that.) I like these changes. Thanks!

confunguido (approved):

PR #125, opened 2025-04-30
Fixed transmission report

KOVALW:
- Refactored the `record_transmission_event` function and updated the event subscription logic to ensure proper handling of infection data.
- Added a new `transmission_report_name` parameter to specify the output file name for the transmission report.


confunguido (approved):
lgtm

PR #132, opened 2025-05-04
Remove global setting type and add explicit way of setting itinerary setting type ratios. Fixes #130.

### LINKED ISSUE #130: Global settings/itinerary needs a test
Settings.rs does not include a test for the global itinerary. The purpose of global setting isn't clear either, so justification for it and how to use it should be included somewhere.

ChiragKumar9:
- Removed global setting type. Instead, created an ad hoc `HomogeneousMixing` setting type with a homogeneous mixing itinerary for use in tests that require people in a setting (to test infection process and contact selection) and homogeneous mixing (like checking the timing of infection attempts).
- Added a way to explicitly specify itinerary ratios each of the core setting types from the config file.
- Kept a `SplitEvenly` itinerary write function type that specifies a constant ratio agnostic to setting type/id for new settings beyond core setting types (like the homogeneous mixing one). This method is meant to be used in tests and not strictly necessary, but the "shoving a round peg in a square hole" that you have to do otherwise hampers code readability significantly.


src/rate_fns/rate_fn_storage.rs
```diff
@@ -169,7 +169,12 @@ mod tests {
             synth_population_file: PathBuf::from("."),
             transmission_report_name: None,
             settings_properties: vec![],
-            itinerary_fn_type: ItineraryWriteFnType::SplitEvenly,
+            itinerary_fn_type: ItineraryWriteFnType::Split {
```
confunguido, line 172: What happens in this case when everything is zero? Do we throw an error?

confunguido (approved):

src/rate_fns/rate_fn_storage.rs
ChiragKumar9: I like that idea. I added a validation check for making sure that all the ratios are not zero and a test for the check.

I switched to using `SplitEvenly` in tests where the setting does not matter as I was previously using `Split {home: 0.0, school: 0.0, work: 0.0, census_tract: 0.0}` but this now fails.

Feel free to re-review if you would like otherwise I will merge.

Issue #101, opened 2025-04-18
Need to test behavior of model with tests to make sure that transmission is happening in all settings as we expect


confunguido:
In the `infection_propagation_loop module`, add tests to make sure that transmission is happening in settings as we expect. If you have someone who is going to school and home and that person's proportions are 0.0 and 1.0, do we see any infections at school from person?

PR #140, opened 2025-05-12
Separating the infection attempt process

erik-rosenstrom:
This PR addresses issue #101. The infection propagation loop is updated to separate the contents of `schedule_next_forecasted_infection` to first select a setting where the contact will occur. This returns a setting type and setting id. These two values are then passed to the modified `get_contact` method. This method extracts the `TypeId` from the setting type and calls the `get_contact_internal` method as before.  Tests were modified in the setting module to include tests for both selecting the setting and selecting the contact for setting.

### DIFF
--- a/src/infection_propagation_loop.rs
+++ b/src/infection_propagation_loop.rs
@@ -4,6 +4,7 @@ use crate::infectiousness_manager::{
 };
 use crate::parameters::{ContextParametersExt, Params};
 use crate::rate_fns::rate_fn_storage::load_rate_fns;
+use crate::settings::ContextSettingExt;
 use ixa::{
     define_rng, trace, Context, ContextPeopleExt, IxaError, PersonId, PersonPropertyChangeEvent,
 };
@@ -19,9 +20,14 @@ fn schedule_next_forecasted_infection(context: &mut Context, person: PersonId) {
         context.add_plan(next_time, move |context| {
             // TODO<ryl8@cc.gov>: We will choose a setting here
             if evaluate_forecast(context, person, forecasted_total_infectiousness) {
-                if let Some(next_contact) = infection_attempt(context, person) {
-                    trace!("Person {person}: Forecast accepted, infecting {next_contact}");
-                    context.infect_person(next_contact, Some(person));
+                if let Some((setting_type, setting_id)) = context.get_setting_for_contact(person) {
+                    trace!("Person {person}: Forecast accepted, setting {setting_id}");
+                    if let Some(next_contact) =
+                        infection_attempt(context, person, setting_type, setting_id)
+                    {
+                        trace!("Person {person}: Forecast accepted, infecting {next_contact}");
+                        context.infect_person(next_contact, Some(person));
+                    }
                 }
             }
             // Continue scheduling forecasts until the person recovers.
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -29,6 +29,7 @@ pub trait SettingType {
         members: &[PersonId],
         setting_properties: SettingProperties,
     ) -> f64;
+    fn get_type_id(&self) -> TypeId;
 }
 
 #[derive(Debug, Hash, Eq, PartialEq, Ord, PartialOrd)]
@@ -174,6 +175,9 @@ macro_rules! define_setting_type {
                 #[allow(clippy::cast_precision_loss)]
                 ((n_members - 1) as f64).powf(setting_properties.alpha)
             }
+            fn get_type_id(&self) -> std::any::TypeId {
+                std::any::TypeId::of::<$name>()
+            }
         }
     };
 }
@@ -213,10 +217,11 @@ pub trait ContextSettingExt {
     ) -> Option<&Vec<PersonId>>;
     fn calculate_total_infectiousness_multiplier_for_person(&self, person_id: PersonId) -> f64;
     fn get_itinerary(&self, person_id: PersonId) -> Option<&Vec<ItineraryEntry>>;
-    fn get_contact<T: SettingType + 'static, Q: Query + 'static>(
+    fn get_contact<Q: Query + 'static>(
         &self,
         person_id: PersonId,
-        setting_id: SettingId<T>,
+        setting_type: &dyn SettingType,
+        setting_id: usize,
         q: Q,
     ) -> Result<Option<PersonId>, IxaError>;
     fn draw_contact_from_transmitter_itinerary<Q: Query>(
@@ -224,6 +229,7 @@ pub trait ContextSettingExt {
         person_id: PersonId,
         q: Q,
     ) -> Result<Option<PersonId>, IxaError>;
+    fn get_setting_for_contact(&self, person_id: PersonId) -> Option<(&dyn SettingType, usize)>;
 }
 
 trait ContextSettingInternalExt {
@@ -448,13 +454,15 @@ impl ContextSettingExt for Context {
             .get(&person_id)
     }
 
-    fn get_contact<T: SettingType + 'static, Q: Query + 'static>(
+    fn get_contact<Q: Query + 'static>(
         &self,
         person_id: PersonId,
-        setting_id: SettingId<T>,
+        setting_type: &dyn SettingType,
+        setting_id: usize,
         q: Q,
     ) -> Result<Option<PersonId>, IxaError> {
-        self.get_contact_internal(person_id, TypeId::of::<T>(), setting_id.id, q)
+        // let container: &SettingDataContainer = self.get_data_container(SettingDataPlugin).unwrap();
+        self.get_contact_internal(person_id, setting_type.get_type_id(), setting_id, q)
     }
     fn draw_contact_from_transmitter_itinerary<Q: Query>(
         &self,
@@ -480,6 +488,30 @@ impl ContextSettingExt for Context {
             )
         } else {
             Ok(None)
+        }
+    }
+    fn get_setting_for_contact(&self, person_id: PersonId) -> Option<(&dyn SettingType, usize)> {
+        let container = self.get_data_container(SettingDataPlugin).unwrap();
+        let mut itinerary_multiplier = Vec::new();
+        container.with_itinerary(person_id, |setting_type, setting_props, members, ratio| {
+            let multiplier = setting_type.calculate_multiplier(members, *setting_props);
+            itinerary_multiplier.push(ratio * multiplier);
+        });
+
+        let setting_index = self.sample_weighted(SettingsRng, &itinerary_multiplier);
+
+        if let Some(itinerary) = self.get_itinerary(person_id) {
+            let itinerary_entry = &itinerary[setting_index];
+            Some((
+                container
+                    .setting_types
+                    .get(&itinerary_entry.setting_type)
+                    .unwrap()
+                    .as_ref(),
+                itinerary_entry.setting_id,
+            ))
+        } else {
+            None
         }
     }
 }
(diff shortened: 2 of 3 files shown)


src/settings.rs
```diff
@@ -482,6 +494,30 @@ impl ContextSettingExt for Context {
             Ok(None)
         }
     }
+    fn get_setting_for_contact(&self, person_id: PersonId) -> Option<(&dyn SettingType, usize)> {
```
ChiragKumar9, line 497: Not for this PR, but as a longer-term discussion, I wonder what the use case is for this function returning None -- won't all people have some itinerary if the itinerary is what sets their setting membership?

src/settings.rs
```diff
@@ -936,6 +972,53 @@ mod test {
...
+        let itinerary_a = vec![
+            ItineraryEntry::new(&SettingId::new(Home, 0), 0.5),
+            ItineraryEntry::new(&SettingId::new(CensusTract, 0), 0.5),
+        ];
+        let itinerary_b = vec![ItineraryEntry::new(&SettingId::new(Home, 0), 1.0)];
+        let _ = context.add_itinerary(person_a, itinerary_a);
+        let _ = context.add_itinerary(person_b, itinerary_b);
+
+        let (setting_type, setting_id) = context.get_setting_for_contact(person_a).unwrap();
+        assert_eq!(setting_type.get_type_id(), TypeId::of::<Home>());
```
ChiragKumar9, line 1009: I would add a comment about why this is always going to be the home setting. Also, the syntax here is going to look a little different once you incorporate changes from #144.

src/infection_propagation_loop.rs
```diff
@@ -393,4 +438,129 @@ mod test {
         // Make sure nothing has happened after person is recovered.
         assert_almost_eq!(context.get_current_time(), recovery_time, 0.0);
     }
+
+    #[test]
+    fn test_location_infections_one_time_unit() {
```
ChiragKumar9, line 443: This test looks good, but I wonder whether you can incorporate it into my test for the infection propagation loop up above -- `test_number_timing_infections_one_time_unit()`.

src/infection_propagation_loop.rs
```diff
@@ -393,4 +439,129 @@ mod test {
...
+        // Does one infectious person generate the number of infections as expected in different
+        // settings? We're going to run many simulations that each start with one infectious and three
+        // susceptible person. Each susceptible person belongs in one of three setting types
+        // and the infectious person is in all three settings. The susceptible people get moved back
+        // to susceptible when becoming infected, so this is really a setup where there is no susceptible
+        // depletion. We stop the simulation at the end of 1.0 time units and compare the number of
+        // infected people in each setting to the expected infectious rate.
+        let num_sims: u64 = 15_000;
```
confunguido, line 452: Is there a reason for 15,000 simulations?

src/infection_propagation_loop.rs
```diff
@@ -393,4 +439,129 @@ mod test {
...
+                    infectious_person,
+                ));
+            }
+            // Add a watcher for when people are infected to record the infection times.
+            context.subscribe_to_event::<PersonPropertyChangeEvent<InfectionStatus>>(
+                move |context, event| {
+                    if event.current == InfectionStatusValue::Infectious {
+                        // Reset the person to susceptible.
+                        if event.person_id == person_home {
+                            *num_infected_home_clone.borrow_mut() += 1;
```
confunguido, line 515: @ChiragKumar9 could you explain to me why we need to count infections like this?

src/infection_propagation_loop.rs
```diff
@@ -393,4 +439,129 @@ mod test {
...
+                            context.set_person_property(
+                                event.person_id,
+                                InfectionData,
+                                InfectionDataValue::Susceptible,
+                            );
+                        } else if event.person_id == person_censustract {
+                            *num_infected_cenustract_clone.borrow_mut() += 1;
```
confunguido, line 522: Why do you need to make people susceptible again instead of running a different simulation?

src/infection_propagation_loop.rs
```diff
@@ -393,4 +439,129 @@ mod test {
...
+            avg_number_infections_home,
+            rate * total_infectiousness_multiplier.unwrap() / 3.0,
+            0.05
+        );
+        #[allow(clippy::cast_precision_loss)]
+        let avg_number_infections_censustract =
+            num_infected_censustract.take() as f64 / num_sims as f64;
+        assert_almost_eq!(
+            avg_number_infections_censustract,
+            rate * total_infectiousness_multiplier.unwrap() / 3.0,
```
confunguido, line 555: can you make this a variable?

src/infection_propagation_loop.rs
ChiragKumar9: There are two things going on here -- a `RefCell` and an `Rc`. Technically, only the `RefCell` part is required, and the `Rc` part adds a guarantee about how our code is working. First, we need the `RefCell` so that we don't have to clone `num_infected_home_clone` each time we bring it into a closure which would result in dropping it once we leave the closure, making is not actually count how many infections we have. Wrapping the `RefCell` in an `Rc` and then cloning adds a guarantee about how the code is working -- it tells us that the only changes being made to our thing inside our `RefCell` are made from our one clone that we know of and there are no other changes being made anywhere (let's say that there was some secret function that could edit what's inside the `RefCell`). So, together, they add validity to how we're testing and basically ensure that we're only counting what we think we're counting and not cloning it all over the place.

src/settings.rs
```diff
@@ -482,6 +494,28 @@ impl ContextSettingExt for Context {
...
+            let itinerary_entry = &itinerary[setting_index];
+            let setting_type = container
+                .setting_types
+                .get(&itinerary_entry.setting_type)
+                .unwrap()
+                .as_ref();
+            Some(SettingId::new(setting_type, itinerary_entry.setting_id))
+        } else {
+            None
+        }
```
confunguido, line 517: Should this be an error?

src/settings.rs
```diff
@@ -936,6 +970,63 @@ mod test {
...
+                },
+            )
+            .unwrap();
+
+        // This section needs to be in its own scope to avoid borrow checker issues.
+        // Basically, adding a new person requires a mutable borrow of context, and calling
+        // methods of setting type requires an immutable borrow of context. Since we add
+        // person c afterwards, this means we have a mutable borrow of context followed by
+        // immutable borrows.
+        {
```
confunguido, line 1001: I'm not sure why you need to separate the first part of the test from the rest.

src/settings.rs
```diff
@@ -964,28 +1055,45 @@ mod test {
...
             .unwrap()
             .is_none());
 
         let person_c = context.add_person(()).unwrap();
-        let itinerary_c = vec![ItineraryEntry::new(&SettingId::new(CensusTract, 0), 0.5)];
-        let _ = context.add_itinerary(person_c, itinerary_c);
+        let itinerary_c = vec![ItineraryEntry::new(SettingId::new(&CensusTract, 0), 0.5)];
+        context.add_itinerary(person_c, itinerary_c).unwrap();
+
+        let temp_census_tract = context
```
confunguido, line 1087: Why do you need to get the reference to the typeId of CensusTract to get contact?

src/transmission_report.rs
```diff
@@ -23,22 +24,37 @@ fn record_transmission_event(
...
+            infection_setting_type,
+            infection_setting_id,
+            ..
+        } = event.current
+        {
+            record_transmission_event(
+                context,
+                event.person_id,
+                infected_by,
+                infection_setting_type.map(std::string::ToString::to_string),
```
confunguido, line 55: This looks very complicated

confunguido (changes requested):
I have some clarifying questions. Please make the requested changes and mark it for review when you're done.

src/settings.rs
```diff
@@ -23,12 +23,14 @@ pub struct SettingProperties {
     pub itinerary_specification: Option<ItinerarySpecificationType>,
 }
 
-pub trait SettingType {
+pub trait SettingType: std::fmt::Debug {
```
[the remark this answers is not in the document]
erik-rosenstrom: It is not necessary to have debug.

src/settings.rs
erik-rosenstrom: Guido also asks if this should be an error rather than returning None. I think since this method is used for infection propagation, we should throw an error here. Presumably we expect everyone to have an itinerary. I will create this error here and it to the test for this method.

src/infection_propagation_loop.rs
erik-rosenstrom: I think it could be incorporated into the test you have written, but I think since it is testing a different mechanism of the model it should be tested separately. If the test fails in the future, I think it will be easier to parse if it is separated. Also, I think Guido made a good point about the test not needing to run for one unit of time, which will give it a slightly different structure.

src/settings.rs
erik-rosenstrom: This is a little more involved than I first thought so I am going to put it is an issue

src/infection_propagation_loop.rs
erik-rosenstrom: You don't need to do this. This has been updated.

src/settings.rs
erik-rosenstrom: I have posted this as an issue. #146

src/settings.rs
erik-rosenstrom: This has been fixed

src/transmission_report.rs
erik-rosenstrom: Cleaned this up.

src/infection_propagation_loop.rs
erik-rosenstrom: The test has been updated to calculate the expected outcome in terms of the itinerary ratios

src/infection_propagation_loop.rs
erik-rosenstrom: There is not a significant reason for this number. It has been reduced to 1000. A large number is generally required to ensure individual simulation stochasticity does not have an impact on the aggregate result.

src/infection_propagation_loop.rs
```diff
@@ -19,9 +20,15 @@ fn schedule_next_forecasted_infection(context: &mut Context, person: PersonId) {
...
-                    trace!("Person {person}: Forecast accepted, infecting {next_contact}");
-                    context.infect_person(next_contact, Some(person));
+                if let Some(setting_id) = context.get_setting_for_contact(person) {
+                    let str_setting = setting_id.setting_type.get_name();
+                    let id = setting_id.id;
+                    if let Some(next_contact) =
+                        infection_attempt(context, person, setting_id)
+                    {
+                        trace!("Person {person}: Forecast accepted, setting type {str_setting} {id}, infecting {next_contact}");
+                        context.infect_person(next_contact, Some(person), Some(str_setting), Some(id));
```
ChiragKumar9, line 30: Nit: just for ease of code reading, I would have `infect_person` take `Option<SettingId>` and then internally in that method it assigns the setting string and id for the person. That way this code stays focused on the logic of how infections happen, not so much the mechanics of what an infection means in terms of person properties.

ChiragKumar9: I think I was the one who wrote `let str_setting = ...` and `let id = ...` up above, but in looking at this again, I think moving the assignment of the string and id could improve readability.

src/infection_propagation_loop.rs
```diff
@@ -393,4 +434,126 @@ mod test {
         // Make sure nothing has happened after person is recovered.
         assert_almost_eq!(context.get_current_time(), recovery_time, 0.0);
     }
+
+    #[test]
+    fn test_location_infections() {
```
ChiragKumar9, line 439: Am I correct in understanding that we only need the new settings you added for this test? If so, I would make a comment for later readability.

src/infection_propagation_loop.rs
```diff
@@ -393,4 +434,126 @@ mod test {
...
+            [0.5, 0.5, 0.0],
+            [0.5, 0.0, 0.5],
+            [0.0, 0.5, 0.5],
+            [0.5, 0.5, 0.5],
+        ];
+        for ratio in ratios {
+            // We need the total infectiousness multiplier for the person.
+            let sum_of_ratio: f64 = ratio.iter().sum();
+            let mut total_infectiousness_multiplier = None;
+            // Where we store the infection times.
```
ChiragKumar9, line 467:
```suggestion
            // Where we store the infection counts.
```

src/infection_propagation_loop.rs
```diff
@@ -393,4 +434,126 @@ mod test {
...
+
+                context.infect_person(infectious_person, None, None, None);
+                // Get the total infectiousness multiplier for comparison to total number of infections.
+                if total_infectiousness_multiplier.is_none() {
+                    total_infectiousness_multiplier = Some(max_total_infectiousness_multiplier(
+                        &context,
+                        infectious_person,
+                    ));
+                }
+                // Add a watcher for when people are infected to record the infection times.
```
ChiragKumar9, line 518:
```suggestion
                // Add a watcher for when people are infected to record their infection setting.
```

src/settings.rs
```diff
@@ -482,6 +494,28 @@ impl ContextSettingExt for Context {
...
+            itinerary_multiplier.push(ratio * multiplier);
+        });
+
+        let setting_index = self.sample_weighted(SettingsRng, &itinerary_multiplier);
+
+        if let Some(itinerary) = self.get_itinerary(person_id) {
+            let itinerary_entry = &itinerary[setting_index];
+            let setting_type = container
+                .setting_types
+                .get(&itinerary_entry.setting_type)
```
ChiragKumar9, line 511: Nit: later, we should audit this code for making sure that when we mean `setting_type` we really mean the setting type and _not_ the type id.

src/settings.rs
ChiragKumar9: Thanks for catching. Once we updated the syntax of `SettingId` to have the right lifetime, we do not need to put this section in its own scope. This has been fixed and the separation and comment removed.

src/infection_propagation_loop.rs
ChiragKumar9: Well, turns out we can't really do this without violating Rust ownership rules. Basically, the setting id already has a reference to the actual setting type stored in it, so we can't them pass it around to a method that takes a mutable reference to context (`infect_person`).

KOVALW (approved):
LGTM before the settings refactor.

PR #144, opened 2025-05-15
Refactor places that require setting location to take `SettingId` rather than a tuple

ChiragKumar9:
Don't fully love that we have to take a reference to a setting when making a new setting id, but this gets the job done of removing the tuple of specifying both `&dyn SettingType` and `id: usize` separately when needing the setting id.

### DIFF
--- a/src/infectiousness_manager.rs
+++ b/src/infectiousness_manager.rs
@@ -8,7 +8,7 @@ use statrs::distribution::Exp;
 use crate::{
     population_loader::Alive,
     rate_fns::{InfectiousnessRateExt, InfectiousnessRateFn, RateFnId, ScaledRateFn},
-    settings::{ContextSettingExt, SettingType},
+    settings::{ContextSettingExt, SettingId, SettingType},
 };
 
 #[derive(Serialize, PartialEq, Debug, Clone, Copy)]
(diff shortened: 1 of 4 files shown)


src/infectiousness_manager.rs
```diff
@@ -71,14 +71,13 @@ pub fn max_total_infectiousness_multiplier(context: &Context, person_id: PersonI
 define_rng!(ForecastRng);
 
 // Infection attempt function for a context and given `PersonId`
-pub fn infection_attempt(
+pub fn infection_attempt<T: SettingType + ?Sized>(
```
erik-rosenstrom, line 74: One thing I got confused about when trying something like this the first time was, was when the `<T: SettingType + etc>` can stop cascading upwards in the code. In my first attempt for example `infection_attempt` had this turbo fish and that seemed to require functions in the infection propagation loop to have the turbo fish.

src/settings.rs
```diff
@@ -33,19 +32,16 @@ pub trait SettingType: std::fmt::Debug {
     fn get_name(&self) -> &'static str;
 }
 
-#[derive(Debug, Hash, Eq, PartialEq, Ord, PartialOrd)]
```
erik-rosenstrom, line 36: Why were the derive properties removed? Were they needed in the first place?

erik-rosenstrom (approved):
I think this is nice. I had two questions, but I think this would be great to merge into my branch. Thank you!

src/infectiousness_manager.rs
ChiragKumar9: It's an excellent question. What I'm saying here in this code is that `T` must implement the `SettingType` trait and may or may not be `Sized`, but you'll notice that in the function signature itself I say that `setting_id` is of type `SettingId<T>`. What this means in this case is that `T` is being inferred from whatever `SettingId` I feed into the function. So, I actually don't need to call the function with the turbofish because `T` can be inferred from function signature.

I'd need to continue propagating the turbofish if this weren't the case. If I recall, your initial code was generic where `T` was being specified in the turbofish, but there were no arguments to let the function figure out what `T` was without being told explicitly, requiring you to continue propagating up the turbofish. This is by no means a hard rule, but in general, most code we end up writing doesn't actually require the turbofish because we just use `T` as a function argument.

src/settings.rs
ChiragKumar9: Yep they weren't necessary! I think they were introduced potentially when `SettingId`s were put in a hashmap because they all have that flavor, but the bottom line is that they aren't necessary.

PR #147, opened 2025-05-20
Enable modular specification of aggregator functions. Fixes #143 and closes #145.

### LINKED ISSUE #143: Improved methods for calling transmission aggregator functions
**Current implementation**
We allow for only one transmission aggregator function per infection status, which inflexibly aggregates all of the floats returned by current registered modifier functions of that infection status. If registering an aggregator that is not the default (i.e. does not simply return the product of all floats), the aggregator must also be generic or must be applied using a `match` statement such as

```rust
for (t, f) in modifier_vector {
    let agg_contrib = match t {
        some_property_type_id => some_fn(f(context, person_id), vec),
        _ => f(context, person_id)
    }
    ...
}
```

which can be cumbersome and will not be modular.

**Desired functionality**
Implement some `Aggregate` method for modifier functions that are known to the user and allow for multiple aggregation methods to accumulate. This could be accomplished similarly to the implementation of `Query`, that can implement `Query` for multiple chained `Query`s. Once this is accomplished, a default aggregator method on the overall floats returned from modifier dependency  relationships could be defined and default to a product.

### LINKED ISSUE #145: How should we identify transmission modifiers?
Right now, we allow for two interfaces to specify transmission modifiers -- as a generic function of `Context` and `PersonId` that returns a float, a blanket catch-all for transmission modifiers that enables the user to specify a transmission modifier that depends on any public methods, and a more specialized interface that deals with the very common use case of specific person property values being associated with a given transmission modifier value.

We need a way to identify transmission modifiers so that we can specify aggregator functions that depend on the transmission in question. Person properties are a natural choice to identify transmission modifiers in the second use case, but they are not a logical choice for identifying the transmission modifier in the first use case where we specify a more generic function that doesn't explicitly depend on a given person property.

Right now, we specify a person property -- even though it is an irrelevant property -- when specifying a general transmission modifier function. This is confusing to the user. There are a few options:

1. We create a new trait called `TransmissionModifier` that is an empty trait. We provide a blanket implementation of `TransmissionModifier` for all person properties, and then we use types that implement `TransmissionModifier` as the identifier for their associated transmission modifier. For the second use case of feeding in a map of how person property values are associated with particular transmission modifier values, the person property still serves as the identifier. For the first use case where we need to feed in a generic function, we just create a unit struct that implements the `TransmissionModifier` trait and use that as the identifier. Under this formulation, the `Aggregate` trait described in #143 would instead be implemented on tuples of `(TransmissionModifier, f64)` rather than `(PersonProperty, f64)`. This does not solve the problem of needing to globally know
2. Instead of storing the transmission modifiers in a hashmap indexed by `TypeId`, which requires an identifier, what if we instead stored just the modifier functions in a vector? To specify an aggregator, we would specify some rules for how this particular intervention is aggregated with others, perhaps through an enum which has variants of `Independent` and `Custom(Box<AggregatorFn>)` though here the `AggregatorFn` will be just a function of the current interventions and the so-far aggregated interventions.

I'm leaning towards the first as being the better solution, but we still need a way to modularly specify transmission modifier aggregators, and the "rules" approach described in the latter sounds better.

CC @KOVALW.

ChiragKumar9:
This PR proposes a way to improve registering a custom aggregator function for transmission modifiers. In particular, the challenge with the old API (#120) was that in order to specify a custom aggregator function (i.e., a function that describes how to turn the modifier values from each transmission intervention into one overall float that is the person's total reduction in infectiousness/susceptibility), the user had to specify a function that depended on all _all_ transmission modifiers. This structure is not very modular.

Instead, this PR proposes an API that lets the user specify aggregator functions that depend on only a subset of transmission modifiers. Users get the value for a particular transmission modifier through the new `evaluate_transmission_modifier` method and combine modifiers in an arbitrary way, returning a float. For all modifiers not included through custom aggregation, the default aggregator (simple multiplication of modifiers) is used. In making these changes, I also had to implement a solution that addresses #145 by introducing a new trait that marks transmission modifiers -- `pub trait TransmissionModifier`.

In the future, we might add yet another degree of generality by allowing the user to override multiplication of modifiers as being the default aggregator in a simple one-line fashion, if there is a use-case for this.

Disclaimer: What I have done here is just a proposal. I encourage anyone interested to noodle around and play with the features they would desire in the API. I am sure this is also not the only way to fix the issues raised in #143 and #145.

### DIFF
--- a/src/interventions/transmission_modifier_manager.rs
+++ b/src/interventions/transmission_modifier_manager.rs
@@ -19,25 +19,6 @@ struct TransmissionModifierContainer {
     modifier_aggregator: HashMap<InfectionStatusValue, Box<TransmissionAggregatorFn>>,
 }
 
-impl TransmissionModifierContainer {
-    fn run_aggregator(
-        &self,
-        context: &Context,
-        person_id: PersonId,
-        infection_status: InfectionStatusValue,
-        modifiers: &[(TypeId, f64)],
-    ) -> f64 {
-        self.modifier_aggregator.get(&infection_status).map_or_else(
-            || default_aggregator(modifiers),
-            |agg| agg(context, person_id),
-        )
-    }
-}
-
-fn default_aggregator(modifiers: &[(TypeId, f64)]) -> f64 {
-    modifiers.iter().map(|&(_, f)| f).product()
-}
-
 define_data_plugin!(
     TransmissionModifierPlugin,
     TransmissionModifierContainer,
@@ -221,12 +202,15 @@ impl ContextTransmissionModifierExt for Context {
                     registered_modifiers.push((*t, f(self, person_id)));
                 }
 
-                transmission_modifier_plugin.run_aggregator(
-                    self,
-                    person_id,
-                    infection_status,
-                    &registered_modifiers,
-                )
+                // Run the aggregator function, defaulting to the default (independence of all
+                // modifiers, a simple product) if no aggregator has been registered
+                transmission_modifier_plugin
+                    .modifier_aggregator
+                    .get(&infection_status)
+                    .map_or_else(
+                        || default_aggregator(&registered_modifiers),
+                        |agg| agg(self, person_id),
+                    )
             } else {
                 // If the infection status is not found in the map, return 1.0
                 1.0
@@ -236,6 +220,10 @@ impl ContextTransmissionModifierExt for Context {
             1.0
         }
     }
+}
+
+fn default_aggregator(modifiers: &[(TypeId, f64)]) -> f64 {
+    modifiers.iter().map(|&(_, f)| f).product()
 }
 
 #[cfg(test)]
--- a/src/interventions/transmission_modifier_manager.rs
+++ b/src/interventions/transmission_modifier_manager.rs
@@ -6,12 +6,13 @@ use std::{any::TypeId, collections::HashMap};
 use crate::infectiousness_manager::{InfectionStatus, InfectionStatusValue};
 
 type TransmissionModifierFn = dyn Fn(&Context, PersonId) -> f64;
-type TransmissionAggregatorFn = dyn Fn(&Vec<(TypeId, f64)>) -> f64;
+type TransmissionAggregatorFn = dyn Fn(&Context, PersonId) -> f64;
 
 pub trait TransmissionModifier: std::fmt::Debug + 'static {}
 
 impl<T> TransmissionModifier for T where T: PersonProperty + std::fmt::Debug + 'static {}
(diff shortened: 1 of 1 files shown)


RobertJacobsonCDC (changes requested):
Instead of making the container or API aware of the dependencies or trying to encode them into the type system, why not just have the modifiers/aggregate functions tell you themselves when they apply?

```rust
pub struct Modifier {
  name: "Example", // debugging purposes
  applies: Fn(&Context, &PersonId) -> bool,
  compute: Fn(&Context, &PersonId) -> f64,
}
```

Then to compute the "modified relative total infectiousness" you go through the list of all registered modifiers and apply the ones that report they should be applied.

(Strictly speaking, we *could* have the modifier itself be responsible for determining if it applies within the compute function, with it's `compute()` doing nothing if it doesn't apply. Then we would just call `modifier.compute()` for every modifier unconditionally. But this separation makes that expectation explicit.)

Similarly for aggregate functions.

RobertJacobsonCDC:
I _think_ I understand a little bit better. Some thoughts:

>  the user had to specify a function that depended on all all transmission modifiers. This structure is not very modular.

No, they only need to specify a function that takes a vector and picks out the values it needs, which is different!

Modifiers that depend on multiple properties simultaneously can be implemented as aggregators—or, since modifiers take a vector of all the (property, value) pairs, such a modifier need only be registered for a single one of its property dependencies.

A modifier that doesn't depend on any property is an aggregator.

The more I think about this, the more I feel like all of this can be collapsed into a single conceptual thing, namely a function that just mutates a vector of `(Option<PropertyId>, f64)` pairs, returning nothing. If it is a modifier that depends on multiple properties, it can select out the whichever pairs it needs to compute with. Then the modifier could either replace those pairs or remove them and append a `(None, f64)` value to the list or leave the pairs and just append a value—depending on your use case/desired behavior. A modifier that depends on no properties can just append a `(None, f64)` value. It's not clear to me what use case there is for an aggregator if you also have modifiers that can depend on multiple properties and mutate the list. I guess you would want to account for dependencies—which we absolutely can do if we need to—but maybe you can get away with just executing modifiers that depend on a single property first.

ChiragKumar9:
Closing this in favor of the API #148 that addresses [1744a63](https://github.com/CDCgov/ixa-epi-isolation/pull/147/commits/1744a634c5f53726bf50fdbb43b236739518a9e3) and [51c8970](https://github.com/CDCgov/ixa-epi-isolation/pull/147/commits/51c89700eb9883c0b5958b9c16513da43daf3dda).

PR #148, opened 2025-05-22
Enable custom specification of arbitrary transmission modifiers. Fixes #145.

### LINKED ISSUE #145: How should we identify transmission modifiers?

ChiragKumar9:


src/interventions/transmission_modifier_manager.rs
```diff
@@ -75,48 +67,38 @@ pub trait ContextTransmissionModifierExt {
...
+        let boxed_transmission_modifier = Box::new(transmission_modifier);
 
         // Insert the boxed function into the transmission modifier map, using entry to handle unititialized keys
         if let Some(_modifier_fxn) = self
             .get_data_container_mut(TransmissionModifierPlugin)
             .transmission_modifier_map
             .entry(infection_status)
             .or_default()
-            .insert(TypeId::of::<T>(), boxed_fn)
+            .insert(TypeId::of::<T>(), boxed_transmission_modifier)
```
KOVALW, line 95: It's a little strange that for store_values type modifiers we now have `HashMap<InfectionStatus, HashMap<T, (T, T::Value)>>`, but I see why this is a convenient API to use

src/interventions/transmission_modifier_manager.rs
```diff
@@ -75,48 +67,38 @@ pub trait ContextTransmissionModifierExt {
...
-    ) where
-        F: Fn(&Context, PersonId) -> f64 + 'static,
-    {
-        // Box the function to store it in the map
-        let boxed_fn = Box::new(modifier_fn);
+        transmission_modifier: T,
+    ) {
+        // Box the transmission modifier to store it in the map
+        // Transmission modifiers must implement debug so that we can more easily log their addition
+        let name = format!("{transmission_modifier:?}");
```
KOVALW, line 86: It doesn't much matter besides for Debugging because the name is unique, but this won't supply the person property as a name, it'll present the entire tuple with the HashMap. I would suggest supplying `name` as a component of the trait and returning `TypeId::of::<T>` as the modifier name, but that would place the burden on the user.

src/interventions/transmission_modifier_manager.rs
```diff
@@ -144,33 +126,10 @@ impl ContextTransmissionModifierExt for Context {
...
-        self.register_transmission_modifier_fn(
-            infection_status,
-            person_property,
-            move |context: &Context, person_id: PersonId| -> f64 {
-                let property_val = context.get_person_property(person_id, person_property);
-                // Return the corresponding value from the map, or 1.0 if not found
-                *modifier_map.get(&property_val).unwrap_or(&1.0)
-            },
-        );
+        self.register_transmission_modifier_fn(infection_status, (person_property, modifier_map));
```
KOVALW, line 129: Slick! I like that the move is no longer necessary.

src/interventions/transmission_modifier_manager.rs
```diff
@@ -331,12 +292,23 @@ mod test {
         assert_eq!(modifier_map.len(), 1);
 
         let modifier_fn = modifier_map
-            .get(&TypeId::of::<MandatoryInterventionStatus>())
+            .get(&TypeId::of::<(
```
KOVALW, line 295: This gets at my comment above, for long lists of key value pairs this TypeId will be incredibly onerous if we ever need to call it. For tests this is fine, but I do lean towards asking the user to define a name and supplying the name method above as I mentioned using `TypeId::of::<T>`

src/interventions/transmission_modifier_manager.rs
```diff
@@ -456,6 +439,23 @@ mod test {
         }
     }
 
+    #[derive(Debug)]
```
KOVALW, line 442: This implementation is great! Much clearer to separate out the trait and not need the closure

KOVALW (commented):
This is a wonderful set of changes. My only concerns are over name identifiers for bookkeeping.

src/interventions/transmission_modifier_manager.rs
RobertJacobsonCDC: There is `std::any::type_name::<T>()` and `std::any::type_name_of_value(T)`, too.

KOVALW (approved):
LGTM! The user shouldn't have to frequently use the `PersonPropertyModifier` syntax under `TypeId`, but maybe eventually we provide something a little more ergonomic

Issue #146, opened 2025-05-19
Add error if getting itinerary is none during infection propagation


erik-rosenstrom:
In the `get_setting_for_contact` method if a person does not have an itinerary, an error should be thrown indicating such. Ideally, since `get_itinerary` returns an option, if it is None an error could be thrown. However, the `with_itinerary` used to generate weights for sampling, and it is this sampling that would break the code if an individual did not have an itinerary. This issue of where the error should be place in the code needs to be addressed. Should the error be moved into the `get_itinerary` method or the `with_itinerary` method?

ChiragKumar9:
The real design question here is whether it is possible for an individual to not have an itinerary. In that case, `get_itinerary` cannot return an option but just an itinerary.

I would argue that all individuals must have an itinerary, and so itineraries should be added with person creation (if we choose to go continue the design we have now where itineraries define setting membership), or itineraries are lazily created when invoking an individual's settings (see #139 for details).

ChiragKumar9:
This is cropping up again in #149 where we are erroring when a person doesn't have an itinerary.

PR #149, opened 2025-05-22
Itinerary modifiers closes #118

### LINKED ISSUE #118: Modifiers on itineraries
To model isolation, we need to address the following: 
- Modify itineraries 
-  Differentiate current and maximum total infectiousness
- How to model the impact of isolation on household contacts.

confunguido:
This is a first attempt to add itinerary modifiers. It introduces a new hashmap of `modified_itineraries` which if present would be used as the itinerary for a person. 

Issues:
- Adding and modifying itineraries implies removing and adding people to a vector. This can be intensive.   
- Right now, ratios are not redistributed among other settings, that is up to the user when defining a modified itinerary.

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -224,6 +224,11 @@ pub trait ContextSettingExt {
         person_id: PersonId,
         itinerary: Vec<ItineraryEntry>,
     ) -> Result<(), IxaError>;
+    fn limit_itinerary_by_setting_types<T: SettingType + 'static>(
+        &mut self,
+        person_id: PersonId,
+        _setting_type: T,
+    ) -> Result<(), IxaError>;
     fn remove_modified_itinerary(
         &mut self,
         person_id: PersonId,
@@ -357,6 +362,33 @@ impl ContextSettingExt for Context {
             .setting_properties
             .insert(TypeId::of::<T>(), setting_props);
         Ok(())
+    }
+
+    // TODO: Limit itinerary to a vector of setting types instead of one setting type
+    fn limit_itinerary_by_setting_types<T: SettingType + 'static>(
+        &mut self,
+        person_id: PersonId,
+        _setting_type: T,
+    ) -> Result<(), IxaError> {
+        let container = self.get_data_container_mut(SettingDataPlugin);
+        match container.itineraries.get(&person_id) {
+            None => Err(IxaError::from("Can't find itinerary for person")),
+            Some(itineraries) => {
+                let mut modified_itinerary = Vec::<ItineraryEntry>::new();
+                for entry in itineraries.iter() {
+                    if entry.setting_type == TypeId::of::<T>() {
+                        modified_itinerary.push(*entry);
+                    }
+                }
+                if modified_itinerary.is_empty() {
+                    return Err(IxaError::from("limit itinerary resulted in empty modified itinerary"));
+                }
+                println!("LIMIT ITINERARY: {:?}", modified_itinerary);
+                
+                self.add_modified_itinerary(person_id, modified_itinerary)?;
+                Ok(())                
+            }
+        }        
     }
 
     fn remove_modified_itinerary(
@@ -1105,12 +1137,9 @@ mod test {
         assert_eq!(h1_members.len(), 3);
         assert_eq!(w_members.len(), 4);
         assert_eq!(s_members.len(), 1);
-        println!("WORK MEMBERS (isolation): {:?}", w_members);
-
 
         let inf_multiplier = context.calculate_total_infectiousness_multiplier_for_person(person_0.unwrap());
-        let expected_multiplier = (1.0) * (2_f64).powf(alpha_h);        
-        println!("MULTIPLIER (isolation): {:?} -- expected {:?}", inf_multiplier, expected_multiplier);
+        let expected_multiplier = (1.0) * (2_f64).powf(alpha_h);
         assert_almost_eq!(inf_multiplier, expected_multiplier, 0.0);
         
         // 3. Remove modified itinerary; get back to normal
@@ -1131,16 +1160,113 @@ mod test {
         assert_eq!(h1_members.len(), 3);
         assert_eq!(w_members.len(), 5);
         assert_eq!(s_members.len(), 2);
-        println!("WORK MEMBERS (post-isolation): {:?}", w_members);
         
         let inf_multiplier = context.calculate_total_infectiousness_multiplier_for_person(person_0.unwrap());
         let expected_multiplier = (1.0/3.0) * (2_f64).powf(alpha_h) +
             (1.0/3.0) * (4_f64).powf(alpha_w) +
             (1.0/3.0) * (1_f64).powf(alpha_s);
-        println!("MULTIPLIER (post-isolation): {:?} -- expected {:?}", inf_multiplier, expected_multiplier);       
+        
         assert_almost_eq!(inf_multiplier, expected_multiplier, 0.0);
     }
 
+        #[test]
+    fn test_limited_itinerary_modifier() {
+        /* H(0) = [0, 1, 2]
+           W(0) = [0,3,4] - 
+          Person 0 isolates using limited itinerary by setting type. Hence, members of 
+         */
+        let mut context = Context::new();
+        context.init_random(42);
+
+        context
+            .register_setting_type(
+                Home,
+                SettingProperties {
+                    alpha: 0.1,
+                    itinerary_specification: None,
+                },
+            )
+            .unwrap();
+        context
+            .register_setting_type(
+                Workplace,
+                SettingProperties {
+                    alpha: 0.3,
+                    itinerary_specification: None,
+                },
+            )
+            .unwrap();
+
+        let person = context.add_person(()).unwrap();
+        let itinerary = vec![
+            ItineraryEntry::new(&SettingId::new(Home, 0), 1.0),
+            ItineraryEntry::new(&SettingId::new(Workplace, 0), 1.0),
+        ];
+        
+        let _isolation_itinerary = [
+            ItineraryEntry::new(&SettingId::new(Home, 0), 1.0),
+        ];
+        
+        let _ = context.add_itinerary(person, itinerary.clone());
+
+        for _ in 0..2 {
+            let itinerary_home = vec![
+            ItineraryEntry::new(&SettingId::new(Home, 0), 1.0),
+            ];
+
+            let p_id = context.add_person(()).unwrap();
+            let _ = context.add_itinerary(p_id, itinerary_home);                        
+        }
+        for _ in 0..2 {
+            let itinerary_work = vec![
+                ItineraryEntry::new(&SettingId::new(Workplace, 0), 1.0),
+            ];
+
+            let p_id = context.add_person(()).unwrap();
+            let _ = context.add_itinerary(p_id, itinerary_work);                        
+        }
+        // Check membership
+        let h_members = context
+            .get_setting_members::<Home>(SettingId::new(Home, 0))
+            .unwrap();
+        let w_members = context
+            .get_setting_members::<Workplace>(SettingId::new(Workplace, 0))
+            .unwrap();
+        assert_eq!(h_members.len(), 3);        
+        assert_eq!(w_members.len(), 3);
+        println!("HOME MEMBERS (limit default): {:?}", h_members);
+        println!("WORK MEMBERS (limit default): {:?}", w_members);
+        
+        // Reduce itinerary to only Home
+        let _ = context.limit_itinerary_by_setting_types(person, Home);
+                // Check membership
+        let h_members = context
+            .get_setting_members::<Home>(SettingId::new(Home, 0))
+            .unwrap();
+        let w_members = context
+            .get_setting_members::<Workplace>(SettingId::new(Workplace, 0))
+            .unwrap();
+        assert_eq!(h_members.len(), 3);        
+        assert_eq!(w_members.len(), 2);
+        println!("HOME MEMBERS (limit isolation): {:?}", h_members);
+        println!("WORK MEMBERS (limit isolation): {:?}", w_members);
+
+        let _ = context.remove_modified_itinerary(person);
+        let h_members = context
+            .get_setting_members::<Home>(SettingId::new(Home, 0))
+            .unwrap();
+        let w_members = context
+            .get_setting_members::<Workplace>(SettingId::new(Workplace, 0))
+            .unwrap();
+        assert_eq!(h_members.len(), 3);        
+        assert_eq!(w_members.len(), 3);
+        
+        println!("HOME MEMBERS (limit isolation): {:?}", h_members);
+        println!("WORK MEMBERS (limit isolation): {:?}", w_members);       
+
+    }
+
+    
     #[test]
     fn test_setting_registration() {
         let mut context = Context::new();
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -371,9 +371,9 @@ impl ContextSettingExt for Context {
         }
         let container = self.get_data_container_mut(SettingDataPlugin);
 
-        // Clean up settings that from previous itinerary, if there is one
-        if let Some(previous_itinerary) = container.modified_itineraries.get(&person_id) {
-            for itinerary_entry in previous_itinerary {
+        // If there's a modified itinerary present, replace with this 
+        if let Some(previous_mod_itinerary) = container.modified_itineraries.get(&person_id) {
+            for itinerary_entry in previous_mod_itinerary {
                 // Remove the person from the previous itinerary
                 container
                     .members
@@ -385,6 +385,22 @@ impl ContextSettingExt for Context {
             }
         }
 
+        // Remove people from default itinerary, if there's one
+        match container.itineraries.get(&person_id) {
+            None => return Err(IxaError::from("Can't modify itinerary if there isn't one present")),
+            Some(previous_itinerary) => {
+                for itinerary_entry in previous_itinerary {
+                    // Remove the person from the previous itinerary
+                    container
+                        .members
+                        .entry(itinerary_entry.setting_type)
+                        .or_default()
+                        .entry(itinerary_entry.setting_id)
+                        .or_default()
+                        .retain(|&x| x != person_id);
+                }
+            }
+        }
         for itinerary_entry in &itinerary {
             // TODO: If we are changing a person's itinerary, the person_id should be removed from vector
             // This isn't the same as the concept of being present or not.
@@ -850,14 +866,13 @@ mod test {
         assert_eq!(members1_removed.len(), 0);
     }
 
-        #[test]
-    fn test_itinerary_modifiers() {
+    #[test]
+    fn test_one_person_itinerary_modifier() {
         /*
         Create two itineraries: default (home, work, censustract), and modified (home)
-        Add one person to the simulation, modify its itinerary
+        Add one person to the simulation, modify its itinerary and check membership
          */
         let mut context = Context::new();
-        context.init_random(1234);
         context
             .register_setting_type(
                 Home,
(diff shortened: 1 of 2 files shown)


src/settings.rs
```diff
@@ -343,6 +512,95 @@ impl ContextSettingExt for Context {
...
+                    if entry.setting_type == setting_type.get_type_id() {
+                        return Ok(Some(entry.setting_id));
+                    }
+                }
+                Ok(None)
+            }
+        }
+    }
+
     fn add_itinerary(
```
KOVALW, line 604: Could you have this call a helper internal function that accepts a HashMap to add the itinerary to and validates it, then calls that function again for adding a modified itinerary?

src/settings.rs
```diff
@@ -302,6 +347,130 @@ impl ContextSettingInternalExt for Context {
         self.get_data_container(SettingDataPlugin)?
             .get_setting_members(&setting_type, setting_id)
     }
+
+    fn add_modified_itinerary(
```
KOVALW, line 351: This should share some common helper functions with `add_itinerary` as I describe below

src/settings.rs
```diff
@@ -211,7 +230,18 @@ pub trait ContextSettingExt {
...
         itinerary: Vec<ItineraryEntry>,
     ) -> Result<(), IxaError>;
+    fn modify_itinerary(
+        &mut self,
+        person_id: PersonId,
+        itinerary_modifier: ItineraryModifiers,
+    ) -> Result<(), IxaError>;
+    fn remove_modified_itinerary(&mut self, person_id: PersonId) -> Result<(), IxaError>;
     fn validate_itinerary(&self, itinerary: &[ItineraryEntry]) -> Result<(), IxaError>;
+    fn get_setting_id(
```
KOVALW, line 240: This isn't really describing what's happening inside the function (getting an id for the first setting that matches a setting type within a person's itinerary), so I might go with a different name. 

I'm also not understanding when this would be called.  How would this be used if the modifier helper functions called by `Isolate` and `Exclude` only use setting types?

src/settings.rs
```diff
@@ -343,6 +512,95 @@ impl ContextSettingExt for Context {
...
+        &mut self,
+        person_id: PersonId,
+        setting_type: &dyn SettingType,
+    ) -> Result<Option<usize>, IxaError> {
+        let container = self.get_data_container_mut(SettingDataPlugin);
+        match container.itineraries.get(&person_id) {
+            None => Err(IxaError::from("Can't find itinerary for person")),
+            Some(itineraries) => {
+                for entry in itineraries {
```
KOVALW, line 594: This will always return the first setting id of a particular setting type. What if someone has multiple workplaces or homes?

src/settings.rs
```diff
@@ -302,6 +347,130 @@ impl ContextSettingInternalExt for Context {
...
+        &mut self,
+        person_id: PersonId,
+        setting_type: &dyn SettingType,
+    ) -> Result<(), IxaError> {
+        let container = self.get_data_container_mut(SettingDataPlugin);
+        match container.itineraries.get(&person_id) {
+            None => Err(IxaError::from("Can't find itinerary for person")),
+            Some(itineraries) => {
+                let mut modified_itinerary = Vec::<ItineraryEntry>::new();
+                for entry in itineraries {
```
KOVALW, line 457: If someone is isolating, they wouldn't necessarily limit themselves to one setting _type_ they would limit themselves to one setting _id_. This would have people stay in all of the `Home` settings registered to their itinerary if isolating.

src/settings.rs
```diff
@@ -814,6 +1067,438 @@ mod test {
...
+        assert_eq!(w_members.len(), 5);
+        assert_eq!(s_members.len(), 2);
+
+        let inf_multiplier =
+            context.calculate_total_infectiousness_multiplier_for_person(person_0.unwrap());
+        let expected_multiplier = (1.0 / 3.0) * (2_f64).powf(alpha_h)
+            + (1.0 / 3.0) * (4_f64).powf(alpha_w)
+            + (1.0 / 3.0) * (1_f64).powf(alpha_s);
+
+        assert_almost_eq!(inf_multiplier, expected_multiplier, 0.001);
```
KOVALW, line 1314: Shouldn't these be exactly equal?

src/settings.rs
```diff
@@ -814,6 +1067,438 @@ mod test {
...
+                School,
+                SettingProperties {
+                    alpha: 0.01,
+                    itinerary_specification: None,
+                },
+            )
+            .unwrap();
+    }
+    #[test]
+    fn test_itinerary_modifiers() {
```
KOVALW, line 1206: Nit: `test_itinerary_modifiers_replace`

src/settings.rs
```diff
@@ -814,6 +1067,438 @@ mod test {
...
+        let inf_multiplier =
+            context.calculate_total_infectiousness_multiplier_for_person(person_0.unwrap());
+        let expected_multiplier = (1.0 / 3.0) * (2_f64).powf(alpha_h)
+            + (1.0 / 3.0) * (4_f64).powf(alpha_w)
+            + (1.0 / 3.0) * (1_f64).powf(alpha_s);
+
+        assert_almost_eq!(inf_multiplier, expected_multiplier, 0.001);
+
+        // 2. Isolate person with itinerary [Home 0 , 1.0]
+        let isolation_itinerary = vec![ItineraryEntry::new(SettingId::new(&Home, 0), 1.0)];
```
KOVALW, line 1257: Can you test with a more complicated itinerary since this is `Replace` and not `Isolate`

src/settings.rs
```diff
@@ -814,6 +1067,438 @@ mod test {
...
+            .unwrap();
+        assert_eq!(h_members.len(), 3);
+        assert_eq!(w_members.len(), 3);
+
+        println!("HOME MEMBERS (limit isolation): {h_members:?}");
+        println!("WORK MEMBERS (limit isolation): {w_members:?}");
+    }
+
+    #[test]
+    fn test_exclude_setting_from_itinerary() {
```
KOVALW, line 1407: It would be nice to see tests with multiple homes registered, a test that gets expected errors if home is absent

src/settings.rs
```diff
@@ -814,6 +1067,438 @@ mod test {
...
+            .unwrap();
+
+        assert_eq!(h_members.len(), 1);
+        assert_eq!(w_members.len(), 2);
+    }
+
+    #[test]
+    fn test_itinerary_modifier_enum() {
+        let mut context = Context::new();
+        context
```
KOVALW, line 1122: Why not call `register_test_settings`?

src/settings.rs
```diff
@@ -814,6 +1067,438 @@ mod test {
...
+        let w_members = context
+            .get_setting_members(SettingId::new(&Workplace, w_id.unwrap()))
+            .unwrap();
+
+        assert_eq!(h_members.len(), 1);
+        assert_eq!(w_members.len(), 2);
+    }
+
+    #[test]
+    fn test_itinerary_modifier_enum() {
```
KOVALW, line 1120: This doesn't seem different from the test below

KOVALW (changes requested):
I think in general this gets the job done of accessing a temporarily stored itinerary using several nice definition methods. A lot of my comments are for reducing redundant code and simplifying the command accessible to the user, but I think that they'll serve to make the module more useable.

I think the larger issues are deciding when to throw errors (which might be upstream of the functions you introduce in this PR) and how to handle multiple setting ids of the same setting type in a single itinerary, which we don't assume to be singularly mapped.

src/settings.rs
```diff
@@ -814,6 +1067,438 @@ mod test {
...
+            .unwrap();
+    }
+    #[test]
+    fn test_itinerary_modifiers() {
+        let mut context = Context::new();
+        register_test_settings(&mut context);
+        let itinerary_map = HashMap::from([
+            (
+                0,
+                HashMap::from([("Home", 0), ("Workplace", 0), ("School", 0)]),
```
ChiragKumar9, line 1212: Why not make the Hashmap hold `&Home` rather than `"Home"` and then you can just feed it straight into the itinerary entry down below?

src/settings.rs
```diff
@@ -814,6 +1067,438 @@ mod test {
...
+            )
+            .unwrap();
+
+        let person = context.add_person(()).unwrap();
+        let itinerary = vec![
+            ItineraryEntry::new(SettingId::new(&Home, 0), 1.0),
+            ItineraryEntry::new(SettingId::new(&Workplace, 0), 1.0),
+        ];
+
+        let _isolation_itinerary = [ItineraryEntry::new(SettingId::new(&Home, 0), 1.0)];
```
ChiragKumar9, line 1440:
```suggestion
```

ChiragKumar9 (changes requested):
I think the API is good. I do think the code could be cleaned up for more easily understanding what is going on, so that others can also edit it. However, most importantly, I feel like we need to see an example of the API in use for isolation guidance to just make sure we haven't missed something we need/usage works with Rust quirks, etc.

erik-rosenstrom (changes requested):
I think its looking good! My main question is about partial isolation.

src/settings.rs
```diff
@@ -63,6 +63,13 @@ impl ItineraryEntry {
     }
 }
 
+#[allow(dead_code)]
+pub enum ItineraryModifiers<'a> {
+    Replace { itinerary: Vec<ItineraryEntry> },
+    Isolate { setting: &'a dyn SettingType },
+    Exclude { setting: &'a dyn SettingType },
```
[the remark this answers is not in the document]
confunguido: Agree with Chirag. It's a bit out of scope, but please create an issue for this. You can do that here by getting itinerary and then doing `append_itinerary_entry`. 

One unresolved issue with this idea, and in general, is where to get that 5% from. I think that has to be specified by the user, but unsure.

src/settings.rs
confunguido: Do you think it should return a vector of setting ids? Right now, I'm assuming an individual only has one setting per setting type. So, the idea is that you can ask someone for their home id `get_setting_id(person, &Home)`. 

I was thinking to use this function for record purposes and to build custom itineraries. For instance, if  a person has an itinerary of `Home(0), Workplace(0), Community(0)` and we want to change that so that instead of community, the person goes to work, home, and a pharmacy.  `Home(0), Workplace(0), Pharmacy(0)`

src/settings.rs
confunguido: I changed this function name to Include, instead of Isolate. It doesn't mean isolate from a public health perspective, it means isolate to a setting type (or types after implementing Chirag's suggestion). 

We should first decide what it means to isolate in the model. If it means more settings than Home with a modified itinerary, it would probably be better implemented by `ReplaceWith` option.

src/settings.rs
```diff
@@ -337,7 +360,85 @@ impl ContextSettingExt for Context {
...
+                    .entry(itinerary_entry.setting_id)
+                    .or_default()
+                    .retain(|&x| x != person_id);
+            }
+        }
+
+        container.modified_itineraries.remove(&person_id);
+
+        // Get people back to default itinerary, if there's one
+        match container.itineraries.get(&person_id) {
```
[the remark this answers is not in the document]
confunguido: I would throw an error because everyone should have a default itinerary.

src/settings.rs
```diff
@@ -343,6 +512,95 @@ impl ContextSettingExt for Context {
...
+                    "Can't modify itinerary if there isn't one present",
+                ))
+            }
+            Some(default_itinerary) => {
+                for itinerary_entry in default_itinerary {
+                    if !container
+                        .setting_types
+                        .contains_key(&itinerary_entry.setting_type)
+                    {
+                        return Err(IxaError::from(
```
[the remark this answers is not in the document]
confunguido: contains key returns a boolean, doesn't it?

src/settings.rs
```diff
@@ -343,6 +512,95 @@ impl ContextSettingExt for Context {
...
+                self.limit_itinerary_by_setting_types(person_id, setting)
+            }
+            ItineraryModifiers::Exclude { setting } => {
+                println!("ItineraryModifier::Exclude -- {:?}", setting.get_type_id());
+                self.exclude_setting_from_itinerary(person_id, setting)
+            }
+        }
+    }
+
+    fn get_setting_id(
```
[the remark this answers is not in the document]
confunguido: The point of this function is to get the id for a setting type, like home. If you want to know the specific home a person has, you can try `get_setting_id(person, &Home)`

src/settings.rs
confunguido: I think I already responded to this above. Should we return a vector? I was working under the assumption of one setting per setting type

src/settings.rs
```diff
@@ -814,6 +1067,438 @@ mod test {
...
+        let w_id = context.get_setting_id(person_two, &Workplace).unwrap();
+
+        let h_members = context
+            .get_setting_members(SettingId::new(&Home, h_id.unwrap()))
+            .unwrap();
+        let w_members = context
+            .get_setting_members(SettingId::new(&Workplace, w_id.unwrap()))
+            .unwrap();
+
+        assert_eq!(h_members.len(), 1);
```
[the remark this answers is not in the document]
confunguido: I want to make sure, not only that the id is 0 but that I can use it to get the right number of people in that house.

src/settings.rs
confunguido: I think they should

src/settings.rs
KOVALW: Just an itinerary that makes two replacements rather than one in the itinerary. This sort of itinerary should be created when testing `Isolate`, which I think should be an extra example test

src/settings.rs
```diff
@@ -337,7 +360,85 @@ impl ContextSettingExt for Context {
...
+                }
+                println!("LIMIT ITINERARY: {modified_itinerary:?}");
+
+                self.add_modified_itinerary(person_id, modified_itinerary)?;
+                Ok(())
+            }
+        }
+    }
+
+    fn remove_modified_itinerary(&mut self, person_id: PersonId) -> Result<(), IxaError> {
```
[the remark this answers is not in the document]
KOVALW: Copying this message here for clarity:
I think in general [my comment on line 346](https://github.com/CDCgov/ixa-epi-isolation/pull/149/files#r2112750666) better reflects what I'm getting at - there should be some function that does accept a HashMap and manipulates it, but that function shouldn't necessarily be exposed to the user through `pub`. In other words, the call to `remove_modified_itinerary` would still be the same in that it only requires a `Context` and `PersonId`, but inside the function it would call a function that accepts a `HashMap`, which is reused throughout the code and may even end up being a method in a trait implemented for the data container instead of living in the context trait extension.

src/settings.rs
```diff
@@ -337,7 +360,85 @@ impl ContextSettingExt for Context {
         Ok(())
     }
 
-    fn add_itinerary(
+    // TODO: Limit itinerary to a vector of setting types instead of one setting type
+    fn limit_itinerary_by_setting_types<T: SettingType + 'static>(
```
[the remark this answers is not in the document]
KOVALW: Filed #156

src/settings.rs
```diff
@@ -63,6 +63,13 @@ impl ItineraryEntry {
     }
 }
 
+#[allow(dead_code)]
+pub enum ItineraryModifiers<'a> {
+    Replace { itinerary: Vec<ItineraryEntry> },
```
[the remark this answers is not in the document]
erik-rosenstrom: Additionally, when using replace it is difficult to know what the contents of the individuals itinerary are, since for an itinerary entry the setting information is private. So it is difficult to recreate someone's itinerary with different ratios.

src/settings.rs
```diff
@@ -352,8 +453,41 @@ impl ContextSettingExt for Context {
             entry.ratio /= total_ratio;
         }
         let container = self.get_data_container_mut(SettingDataPlugin);
-        let mut current_setting_ids = vec![];
 
+        // If there's a modified itinerary present, replace with this
+        if let Some(previous_mod_itinerary) = container.modified_itineraries.get(&person_id) {
```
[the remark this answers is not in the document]
confunguido: I implemented this. Thanks.

src/settings.rs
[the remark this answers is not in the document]
erik-rosenstrom: I see. I think the remaining challenge would be knowing what setting types the individual is a part of. Currently that is private from the individual's itinerary entry. It could be possible to modify the get_setting_id to return an optional vector, and iterate safely over each setting type.

src/settings.rs
```diff
@@ -115,36 +125,40 @@ fn get_itinerary_ratio<T: SettingType + 'static>(
...
     }
+    fn get_itinerary(&self, person_id: PersonId) -> Option<&Vec<ItineraryEntry>> {
+        if let Some(modified_itinerary) = self.modified_itineraries.get(&person_id) {
+            return Some(modified_itinerary);
+        }
+
+        if let Some(itinerary) = self.itineraries.get(&person_id) {
+            return Some(itinerary);
+        }
+        None
```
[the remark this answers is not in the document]
erik-rosenstrom: This is related to issue #146

src/settings.rs
```diff
@@ -307,6 +378,113 @@ impl ContextSettingInternalExt for Context {
...
+        }
+        container.modified_itineraries.insert(person_id, itinerary);
+
+        Ok(())
+    }
+
+    fn exclude_setting_from_itinerary(
+        &mut self,
+        person_id: PersonId,
+        setting_type: &dyn SettingType,
```
ChiragKumar9, line 439: Why does this method take `&dyn SettingType` and not `setting_type: impl SettingType`?

src/settings.rs
```diff
@@ -307,6 +378,113 @@ impl ContextSettingInternalExt for Context {
...
+
+                self.add_modified_itinerary(person_id, modified_itinerary)?;
+                Ok(())
+            }
+        }
+    }
+    fn limit_itinerary_by_setting_type(
+        &mut self,
+        person_id: PersonId,
+        setting_type: &dyn SettingType,
```
ChiragKumar9, line 465: Same comment about method signature.

src/settings.rs
ChiragKumar9: Usually, we use `&dyn Trait` when we want to hold the trait object in another object (like `SettingId`) or we only have access to the trait object from something else and the trait object doesn't implement copy.

src/settings.rs
```diff
@@ -302,6 +373,113 @@ impl ContextSettingInternalExt for Context {
...
+        Ok(())
+    }
+
+    fn exclude_setting_from_itinerary(
+        &mut self,
+        person_id: PersonId,
+        setting_type: &dyn SettingType,
+    ) -> Result<(), IxaError> {
+        let container = self.get_data_container_mut(SettingDataPlugin);
+        match container.itineraries.get(&person_id) {
```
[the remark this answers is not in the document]
ChiragKumar9: I don't think so. We want to modify the original itinerary, not the potentially already modified itinerary. The `get_itinerary` method returns the modified itinerary if there is one first.

src/settings.rs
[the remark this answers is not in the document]
ChiragKumar9: Ah, yes -- so instead of `contains_key` use `get(key)`?

src/settings.rs
```diff
@@ -343,6 +521,90 @@ impl ContextSettingExt for Context {
...
+                        .setting_types
+                        .contains_key(&itinerary_entry.setting_type)
+                    {
+                        return Err(IxaError::from(
+                            "Itinerary entry setting type not registered",
+                        ));
+                    }
+
+                    //container.add_member_to_itinerary_setting(person_id, itinerary_entry);
+                    container
```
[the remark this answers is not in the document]
ChiragKumar9: I agree. If it's not needed, please remove commented code.

src/settings.rs
```diff
@@ -348,6 +526,90 @@ impl ContextSettingExt for Context {
...
+    }
+
+    fn modify_itinerary(
+        &mut self,
+        person_id: PersonId,
+        itinerary_modifier: ItineraryModifiers,
+    ) -> Result<(), IxaError> {
+        match itinerary_modifier {
+            ItineraryModifiers::ReplaceWith { itinerary } => {
+                println!("ItineraryModifier::Replace --  {itinerary:?}");
```
ChiragKumar9, line 579: Please use `trace!` instead.

ChiragKumar9: Please also include the person id here for easier debugging.

src/settings.rs
```diff
@@ -343,6 +521,90 @@ impl ContextSettingExt for Context {
...
+            }
+            ItineraryModifiers::Include { setting } => {
+                println!("ItineraryModifier::Include -- {:?}", setting.get_type_id());
+                self.limit_itinerary_by_setting_type(person_id, setting)
+            }
+            ItineraryModifiers::Exclude { setting } => {
+                println!("ItineraryModifier::Exclude -- {:?}", setting.get_type_id());
+                self.exclude_setting_from_itinerary(person_id, setting)
+            }
+        }
```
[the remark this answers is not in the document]
ChiragKumar9: Love this -- for the future. I think trace statements right now deal with the debugging challenges, and I'm not sure we'd need an event for this under the current framework. Agree that this is the right way to go long term and a great idea.

src/settings.rs
```diff
@@ -348,6 +526,90 @@ impl ContextSettingExt for Context {
...
+        setting_type: T,
+    ) -> Vec<usize> {
+        let container = self.get_data_container_mut(SettingDataPlugin);
+        match container.itineraries.get(&person_id) {
+            None => Vec::new(),
+            Some(itineraries) => {
+                let mut setting_id_vec = Vec::new();
+                for entry in itineraries {
+                    if entry.setting_type == setting_type.get_type_id() {
+                        setting_id_vec.push(entry.setting_id);
```
ChiragKumar9, line 605: How does this work?

```suggestion
                        setting_id_vec.push(SettingId::new(entry.setting_type, entry.setting_id));
```

You may need to make `Copy` a supertrait of `SettingType`. I will mess around on locally too if I get a chance. This enables us to change the signature of this method to be `Vec<SettingId<T>>`.

ChiragKumar9 (changes requested):
I understand that you want to merge this in quick, but I fear that a quick merge in means that some issues with the code will never be addressed. I'd like at least the commented code to be removed. If some of the other syntax changes I've suggested will take too long, we can file issues. In any case, I think there need to be some issues created from the first round of review on this PR.

src/settings.rs
```diff
@@ -307,6 +378,113 @@ impl ContextSettingInternalExt for Context {
...
+            Some(previous_itinerary) => {
+                container
+                    .remove_member_from_itinerary_settings(person_id, previous_itinerary.clone());
+            }
+        }
+
+        for itinerary_entry in &itinerary {
+            // TODO: If we are changing a person's itinerary, the person_id should be removed from vector
+            // This isn't the same as the concept of being present or not.
+            if !container
```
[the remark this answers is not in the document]
erik-rosenstrom: I think it might be better in `validate_itinerary`?

ChiragKumar9 (approved):

PR #151, opened 2025-05-30
Seed the model with recovered individuals. Closes #116.

### LINKED ISSUE #116: Immunity and initial conditions
Two processes need to be incorporated in the model, immunity and initial conditions of such immunity. For instance, at the beginning of the simulation 50% of the population may be immune to the virus due to previous infection. We can model this immunity as permanent or temporary, as well as all-or-none or partial immunity. 

This is issue is two-fold:
1) Decide what the process of immunity looks like
2) Implement initial conditions and immunity if needed

ChiragKumar9:
Proposes an all-or-nothing immunity scheme for a number of individuals to start the simulation as recovered.

If we choose to model partial immunity and certainly if we choose to model continuously waning immunity, I would advocate for us to rephrase our model in an SIS framework where individuals have some degree of protection to infection -- registered in as a transmission intervention -- when returning to the S compartment. That degree of protection can change, and it is calculated on the fly from the registered transmission intervention function.

### DIFF
--- a/input/input.json
+++ b/input/input.json
@@ -3,7 +3,7 @@
     "max_time": 200.0,
     "seed": 123,
     "infectiousness_rate_fn": {"EmpiricalFromFile": {"file": "input/library_empirical_rate_fns.csv", "scale": 1.0}},
-    "initial_incidence": 0.01,
+    "initial_incidence": 0.1,
     "initial_recovered": 0.35,
     "settings_properties": {"Home": {"alpha": 0.0, "itinerary_specification": {"Constant": {"ratio": 0.25}}},
                             "Workplace": {"alpha": 0.0, "itinerary_specification": {"Constant": {"ratio": 0.25}}},
--- a/src/infection_propagation_loop.rs
+++ b/src/infection_propagation_loop.rs
@@ -170,8 +170,10 @@ mod test {
             InfectionDataValue,
         },
         interventions::ContextTransmissionModifierExt,
-        parameters::CoreSettingsTypes,
-        parameters::{GlobalParams, ItinerarySpecificationType, Params, RateFnType},
+        parameters::{
+            ContextParametersExt, CoreSettingsTypes, GlobalParams, ItinerarySpecificationType,
+            Params, RateFnType,
+        },
         rate_fns::{load_rate_fns, InfectiousnessRateExt},
         settings::{
             init as settings_init, CensusTract, ContextSettingExt, Home, ItineraryEntry, SettingId,
@@ -201,7 +203,7 @@ mod test {
     ) -> Context {
         let mut context = Context::new();
         let parameters = Params {
-            initial_incidence: 0.5, // 50% of the population
+            initial_incidence: 0.1, // 10% of the population
             initial_recovered,
             max_time: 100.0,
             seed,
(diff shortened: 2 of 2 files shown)


KOVALW (approved):
LGTM! I like the closure call for each `seed_fn` that'll give us some great future flexibility

erik-rosenstrom (approved):
Looks good to me!

src/infection_propagation_loop.rs
```diff
@@ -45,45 +45,86 @@ fn schedule_recovery(context: &mut Context, person: PersonId) {
...
+    query_susceptibles_and_seed(context, initial_infections, |context, person_id| {
+        trace!("Infecting person {person_id} as an initial infection.");
+        context.infect_person(person_id, None, None, None);
+    })
+}
+
+fn seed_initial_recovered(context: &mut Context, initial_recovered: usize) -> Result<(), IxaError> {
+    trace!("Seeding initial recovered.");
+    query_susceptibles_and_seed(context, initial_recovered, |context, person_id| {
+        // First immediately infect the person so that the infection data is properly set up
```
eqmooring, line 89: I'm unsure what "infection data" means here. Does this mean that recovered individuals somehow get documented as incident infections at t = 0? If so--and depending on the details of the output reports--I think that could risk errors by analysts.

src/infection_propagation_loop.rs
```diff
@@ -45,45 +45,86 @@ fn schedule_recovery(context: &mut Context, person: PersonId) {
...
+        ..
     } = context.get_params();
 
     load_rate_fns(context)?;
 
-    // Seed the initial population
+    // First we seed recovered people. Since recovering people requires setting them as infectious
+    // first to set the infection data, this will trigger the `PersonPropertyChangeEvent` for people
+    // becoming infectious if we do this after setting up the event subscription. Instead, we do this
+    // first, so it happens silently.
```
eqmooring, line 109: Ah, okay, I think this addresses the concern I raised in the previous comment about how this gets recorded. (I'm still a little unclear on "infection data" in this context.)

eqmooring (approved):
I think that this simple approach to seeding immunity in the S-I-R context makes sense. I made a few comments, but I think they comments are more things I am unsure about, rather than specific pieces of feedback to be addressed.

src/parameters.rs
```diff
@@ -208,7 +211,8 @@ mod test {
     #[test]
     fn test_validate_max_time() {
         let parameters = Params {
-            initial_infections: 1,
+            initial_infections: 0,
```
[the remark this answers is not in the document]
ChiragKumar9: It turns out that the initial infections number in the parameters was never being used in those tests. I thought it was confusing that `initial_infections` was nonzero but never being used, so I changed it to zero.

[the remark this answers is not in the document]
ChiragKumar9: In most cases, we want to add our infections manually in the tests -- "let's add a person, now let's infect them!" rather than calling the machinery in the infection propagation loop init to do it for us based on the value of `initial_infections` in the parameters. To make this explicit, I removed the initial infections as being nonzero. But, I added a comment in to make this more clear.

input/input.json
```diff
@@ -4,6 +4,7 @@
     "seed": 123,
     "infectiousness_rate_fn": {"EmpiricalFromFile": {"file": "input/library_empirical_rate_fns.csv", "scale": 1.0}},
     "initial_infections": 1,
+    "initial_recovered": 10,
```
confunguido, line 7: I think it would be better to provide with a proportion of the population currently immune. I think in general we would like to allow for both infections and recovered to be a number or a proportion. But, for initial infections it is fine to have 1 person. I don't think this is as useful for recovered people as 10 recovered people could be anywhere from 0-100% population recovered.

confunguido (approved):
lgtm

PR #152, opened 2025-05-30
Add support for asymptomatic infections with different rate fns. Closes #133.

### LINKED ISSUE #133: Including asymptomatics in the model
Right now, we are using infectiousness rates from our isolation guidance model -- an analysis that included only symptomatic individuals. We need to decide how we want to model asymptomatic infections in the model/whether we want to include asymptomatic infections in the model. Here are two options for modeling the infectiousness of asymptomatic individuals:

1. We just use the rate functions we have currently been using, potentially upsampling category 3 or 4, assuming that asymptomatic and symptomatic infectiousness are equivalent.
2. We use different rate functions, such as one from the literature or NNH's GI.

In both cases, we need to confirm that we are not accidentally assigning symptoms to asymptomatic individuals, though this becomes straightforward with the ability to specify arbitrary relationships between natural history parameters introduced in #106.

There is also the question of whether we feed asymptomatic rate functions into the model in the same file as we feed symptomatic rate functions (#88), but this is an implementation question where we should just decide on something and see how it goes.

ChiragKumar9:


src/symptom_progression.rs
```diff
@@ -224,11 +225,20 @@ fn event_subscriptions(context: &mut Context) {
...
-                    Some(SymptomValue::Presymptomatic),
-                );
+                // Have we assigned this person to be symptomatic when we assigned their infection
+                // data when infecting them?
+                let InfectionDataValue::Infectious { symptomatic, .. } =
+                    context.get_person_property(event.person_id, InfectionData)
+                else {
+                    panic!("Person {:?} is not infectious", event.person_id)
+                };
+                if symptomatic {
```
erik-rosenstrom, line 235: It might be helpful to have a test that the correct proportion of infections lead to symptoms

src/rate_fns/rate_fn_storage.rs
```diff
@@ -1,58 +1,71 @@
...
-            .as_ref()
+        // Get if the person is symptomatic or not and chose rates accordingly if we have separate
+        // asymptomatic rates loaded
+        let separate_asymptomatic_rates = self.get_params().asymptomatic_rate_fn.is_some();
+        let InfectionDataValue::Infectious { symptomatic, .. } =
+            self.get_person_property(person_id, InfectionData)
+        else {
+            panic!("Person {person_id} is not infectious")
+        };
+        if symptomatic || !separate_asymptomatic_rates {
```
erik-rosenstrom, line 62: Nit: I think we need to call this out in any documentation, that if no asymptomatic rates are provided, the symptomatic ones are used.

src/rate_fns/rate_fn_storage.rs
```diff
@@ -1,58 +1,71 @@
...
+            .get_data_container(RateFnPlugin)
+            .expect("Expected rate functions to be loaded.");
         let id = self.get_parameter_id(RateFn, person_id);
-        self.get_data_container(RateFnPlugin)
-            .expect("Expected rate functions to be loaded.")
-            .rates[id]
-            .as_ref()
+        // Get if the person is symptomatic or not and chose rates accordingly if we have separate
+        // asymptomatic rates loaded
+        let separate_asymptomatic_rates = self.get_params().asymptomatic_rate_fn.is_some();
```
eqmooring, line 56: Minor comment: maybe there's a name for `separate_asymptomatic_rates` that better communicates it's not the rates themselves but rather checking whether they exist? Maybe `asymptomatic_rates_specified`?

eqmooring (approved):
I think this looks good. I wonder if there might be a way to have a more generalized approach to inputting rate functions and symptom parameters / lack of symptoms. But I am not sure what that would be, and I think this is fine for now. Could an infection move from symptomatic to asymptomatic due to treatment (e.g., post-exposure prophylaxis)? I don't think that's a priority capability, but any stretch of the imagination, just flagging it as possibly a theoretical limitation. You start to get at this in the inline comments, but we'll need to be clear in documentation that the example trajectory numbers based on categories 3 and 4 are really beyond the scope of what we were looking at scientifically when we were working to infer those categories.

src/symptom_progression.rs
```diff
@@ -556,4 +569,42 @@ mod test {
...
+        // equal to the fraction of people who are symptomatic.
+        let mut num_people_symptomatic_total = 0;
+        let num_people = 1000;
+        let num_sims = 100;
+        let symptomatic_rate = 0.4; // 40% of people should be symptomatic
+        for seed in 0..num_sims {
+            let mut context = setup(1.0 - symptomatic_rate);
+            context.init_random(seed);
+            // Add our people
+            for _ in 0..num_people {
```
erik-rosenstrom, line 587: Nit: Is it faster to do 100 simulations of 1000 people or one simulation of 100,000 people?

erik-rosenstrom (approved):
Looks good to me!

PR #154, opened 2025-06-02
Updated Isolation Policy

erik-rosenstrom:
This PR implements the updated isolation policy. The updated isolation guidance requires an individual to isolate from symptom onset until their symptoms improve. After they exit isolation, the individual takes post isolation precautions for 5 days.

It uses transmission modifiers to implement isolation and facemask interventions. Five input parameters are added to define the post-isolation duration, the probability of following the isolation guidance conditional on being symptomatic, the maximum policy onset delay (the delay distribution is currently assumed to be a uniform distribution), and the relative transmission multiplier values for isolation and facemasks.   

Complexity arises from interaction between the delay distribution and policy uptake. Currently, if an individual's delay to starting isolation is greater than the length of their symptoms, they will not follow post-isolation precautions. This occurs because of the if statement when starting masking that checks that individual was isolating (i.e., they were selected to follow the policy according to the uptake).


src/parameters.rs
```diff
@@ -61,6 +61,17 @@ pub struct Params {
...
     pub transmission_report_name: Option<String>,
+    /// The duration that an individual will follow post-isolation precautions
+    /// following the updated isolation guidance.
+    pub post_isolation_duration: f64,
+    /// The probability that an individual will follow the updated isolation guidance.
+    pub uptake_probability: f64,
+    /// The maximum delay before an individual will follow the guidance
+    pub maximum_uptake_delay: f64,
+    /// Transmission modifier associated with the facemask intervention
+    pub facemask_transmission_modifier: f64,
```
confunguido, line 72: What does modifier mean? Efficacy? 1 - Efficacy?

src/rate_fns/rate_fn_storage.rs
```diff
@@ -211,6 +211,11 @@ mod tests {
             synth_population_file: PathBuf::from("."),
             transmission_report_name: None,
             settings_properties: HashMap::new(),
+            post_isolation_duration: 0.0,
```
confunguido, line 214: Is there a way to run these tests without specifying these values?

src/updated_guidance_interventions.rs
confunguido, line 1: I don't think this file belongs to this PR.

src/updated_guidance.rs
```diff
@@ -0,0 +1,241 @@
...
+                            .is_some()
+                        {
+                            context.set_person_property(event.person_id, IsolatingStatus, true);
+                            trace!("Person {} is now isolating", event.person_id);
+                        }
+                    });
+                }
+            }
+            None => {
+                if context.get_person_property(event.person_id, IsolatingStatus) {
```
confunguido, line 88: We probably should start using itinerary modifiers here.

confunguido (changes requested):
Looking good. Main changes requested:
1) remove secondary implementation file
2) Can we find a way to not have to specify all parameters in all tests from other modules all the time? 
3) We should start incorporating isolation using itinerary modifiers. Maybe this should be a PR to that branch and not to main.

src/updated_guidance_interventions.rs
erik-rosenstrom: This has been deleted

src/parameters.rs
KOVALW: Agreed that these parameter names are vague

src/rate_fns/rate_fn_storage.rs
KOVALW: That would require having Default implemented for the global params, which I would like to add but I think there was push back on it because it allows things to sneak through implicitly

src/natural_history_parameter_manager.rs
```diff
@@ -338,6 +338,11 @@ mod test {
             synth_population_file: PathBuf::from("."),
             transmission_report_name: None,
             settings_properties: HashMap::new(),
+            post_isolation_duration: 0.0,
```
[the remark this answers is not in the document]
erik-rosenstrom: I have turned the parameters for isolation policy into a hashmap, so the parameters don't need to be explicitly supplied to a test. Instead `isolation_policy_parameters: HashMap::new(),` is supplied

src/updated_guidance.rs
KOVALW: Wouldn't we want to bring Guido's implementation from #149 in here? I think I'm missing something

src/rate_fns/rate_fn_storage.rs
erik-rosenstrom: See the previous comment

src/updated_guidance.rs
```diff
@@ -0,0 +1,241 @@
...
+    post_isolation_duration: f64,
+    uptake_probability: f64,
+    maximum_uptake_delay: f64,
+) {
+    context.subscribe_to_event::<PersonPropertyChangeEvent<Symptomatic>>(move |context, event| {
+        match event.current {
+            Some(false) => (),
+            Some(true) => {
+                if context.sample_bool(PolicyRng, uptake_probability) {
+                    let uniform = Uniform::new(0.0, maximum_uptake_delay).unwrap();
```
KOVALW, line 73: Fine to have this implementation for now, but any particular reason you've gone with a uniform instead of exponetital with average delay rate? That would make more sense to me and be easier to test. I could see situations where you draw a random value greater than the symptomatic period, but it's not like you exclude that possibility here, you just reject the delay

src/updated_guidance.rs
```diff
@@ -0,0 +1,241 @@
...
+        let isolation_transmission_modifier = 0.5;
+        let mut context = setup_context(
+            post_isolation_duration,
+            uptake_probability,
+            maximum_uptake_delay,
+            facemask_transmission_modifier,
+            isolation_transmission_modifier,
+        );
+
+        let start_time_symptoms = Rc::new(RefCell::new(0.0f64));
```
KOVALW, line 178: I'm not quite seeing why you need to declare the parameters like this. Do you need the RefCells and do you need to declare all these things for only one iteration through context?

input/input.json
```diff
@@ -11,6 +11,13 @@
     "symptom_progression_library" : {"EmpiricalFromFile": {"file": "input/library_symptom_parameters.csv"}},
     "report_period": 1.0,
     "synth_population_file": "input/people_test.csv",
-    "transmission_report_name": "transmission_report.csv"
+    "transmission_report_name": "transmission_report.csv",
+    "isolation_policy_parameters": {
+      "post_isolation_duration": 5.0,
+      "uptake_probability": 1.0,
+      "maximum_uptake_delay": 3.0,
```
ChiragKumar9, line 18: How does this stack up compared to the parameters you have in the GCM model? I think there are two philosophies -- let's land a minimal set up parameters that we need for now vs let's try to recreate the GCM parameters. Just curious where this implementation stacks out.

src/updated_guidance.rs
```diff
@@ -0,0 +1,258 @@
...
+use crate::{
+    infectiousness_manager::InfectionStatusValue,
+    interventions::ContextTransmissionModifierExt,
+    parameters::ContextParametersExt,
+    symptom_progression::{SymptomValue, Symptoms},
+};
+
+define_person_property_with_default!(MaskingStatus, bool, false);
+define_person_property_with_default!(IsolatingStatus, bool, false);
+define_derived_property!(Symptomatic, Option<bool>, [Symptoms], |data| match data {
```
ChiragKumar9, line 17:
```suggestion
define_derived_property!(PresentingWithSymptoms, Option<bool>, [Symptoms], |data| match data {
```

because a presymptomatic person will become symptomatic.

ChiragKumar9: Also, why not have this just be a bool? Not sure I see the use case for the `Some(false)` variant.

src/updated_guidance.rs
```diff
@@ -0,0 +1,258 @@
...
+) {
+    context.subscribe_to_event::<PersonPropertyChangeEvent<Symptomatic>>(move |context, event| {
+        match event.current {
+            Some(false) => (),
+            Some(true) => {
+                if context.sample_bool(PolicyRng, uptake_probability) {
+                    let uniform = Uniform::new(0.0, maximum_uptake_delay).unwrap();
+                    let delay_period = context.sample_distr(PolicyRng, uniform);
+                    let t = context.get_current_time();
+                    context.add_plan(t + delay_period, move |context| {
```
ChiragKumar9, line 79:
```suggestion
                    context.add_plan(t + delay_period, |context| {
```

input/input.json
erik-rosenstrom: The GCM model has parameters that describe the probability of isolating conditional on symptoms, the probability of mask wearing conditional on isolating, and the duration of post isolation precautions.

src/updated_guidance.rs
```diff
@@ -0,0 +1,241 @@
...
+            InfectionStatusValue::Infectious,
+            MaskingStatus,
+            &[(true, facemask_transmission_modifier)],
+        )
+        .unwrap();
+    context
+        .store_transmission_modifier_values(
+            InfectionStatusValue::Infectious,
+            IsolatingStatus,
+            &[(true, isolation_transmission_modifier)],
```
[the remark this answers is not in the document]
erik-rosenstrom: This will be done once Guido's itinerary modifiers are merged into main. There is a branch off of Guido's work in progress ready to quickly implement these changes

src/rate_fns/rate_fn_storage.rs
erik-rosenstrom: Parameters are implemented for the policy using an API like `Option<InterventionPolicyParameters>` where `InterventionPolicyParameters` is a structure containing post isolation duration etc. It can be specificed as `None` in the tests

src/updated_guidance.rs
erik-rosenstrom: I think I am going to remove this distribution for now and have the uptake delay be constant and specified by the user. This is more in line with individual level model.

src/updated_guidance.rs
```diff
@@ -0,0 +1,367 @@
...
+            move |context, event| {
+                println!(
+                    "symptom time: {}, event.current: {:?}",
+                    context.get_current_time(),
+                    event.current
+                );
+                match event.current {
+                    Some(false) => (),
+                    Some(true) => {
+                        *start_time_symptoms_clone1.borrow_mut() = context.get_current_time();
```
ChiragKumar9, line 255: Could you address this by instead storing the information in a person property, getting rid of the refcell need? This doesn't really feel like a refcell case since you have effectively one run through of the simulation.

src/updated_guidance.rs
```diff
@@ -0,0 +1,367 @@
...
+        for seed in 0..num_sims {
+            let num_people_isolating_clone = Rc::clone(&num_people_isolating);
+            let mut context = setup_context(
+                post_isolation_duration,
+                uptake_probability,
+                uptake_delay_period,
+                facemask_efficacy,
+            );
+            context.init_random(seed);
+            let _first_person: ixa::PersonId = context.add_person(()).unwrap();
```
ChiragKumar9, line 337:
```suggestion
            context.add_person(()).unwrap();
```

src/updated_guidance.rs
```diff
@@ -0,0 +1,349 @@
...
+
+define_person_property_with_default!(MaskingStatus, bool, false);
+define_person_property_with_default!(IsolatingStatus, bool, false);
+define_derived_property!(
+    PresentingWithSymptoms,
+    Option<bool>,
+    [Symptoms],
+    |data| match data {
+        Some(SymptomValue::Presymptomatic) => Some(false),
+        Some(
```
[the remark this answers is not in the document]
erik-rosenstrom: Going with `Some(_)`

src/updated_guidance.rs
```diff
@@ -0,0 +1,430 @@
...
+        Ok(())
+    }
+
+    fn modify_masking_status(
+        &mut self,
+        person: PersonId,
+        masking_status: bool,
+    ) -> Result<(), IxaError> {
+        self.set_person_property(person, MaskingStatus, masking_status);
+        if masking_status {
```
confunguido, line 74: Masking status should not modify itinerary.

src/updated_guidance.rs
```diff
@@ -0,0 +1,367 @@
...
+}
+
+pub fn init(context: &mut Context) {
+    let &Params {
+        intervention_policy_parameters,
+        facemask_parameters,
+        ..
+    } = context.get_params();
+
+    if let Some(facemask_parameters) = facemask_parameters {
```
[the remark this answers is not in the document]
confunguido: I see what this is doing, but I'm not sure that's the best way. In this case, it could happen that someone specifies facemask parameters and not isolation parameters. If this is the isolation policy manager, doesn't it require to have them all those parameters specified given that facemask is part of the guidance?

input/input.json
```diff
@@ -1,6 +1,6 @@
 {
   "epi_isolation.GlobalParams": {
-    "max_time": 200.0,
+    "max_time": 50.0,
```
confunguido, line 3: why do we need to change this?

src/updated_guidance.rs
```diff
@@ -0,0 +1,430 @@
...
+    fn modify_isolation_status(
+        &mut self,
+        person: PersonId,
+        isolation_status: bool,
+    ) -> Result<(), IxaError> {
+        self.set_person_property(person, IsolatingStatus, isolation_status);
+        if isolation_status {
+            self.modify_itinerary(person, ItineraryModifiers::RestrictTo { setting: &Home })?;
+        } else {
+            self.remove_modified_itinerary(person)?;
```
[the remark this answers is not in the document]
erik-rosenstrom: I see what you are saying. I have made the change you suggested. I think this should ultimately be handled by an intervention manager. I will add this as an issue.

src/updated_guidance.rs
```diff
@@ -0,0 +1,394 @@
...
+    ) {
+        if self.sample_bool(
+            PolicyRng,
+            intervention_policy_parameters.isolation_probability,
+        ) {
+            self.add_plan(
+                self.get_current_time() + intervention_policy_parameters.isolation_delay_period,
+                move |context| {
+                    if context.get_person_property(person_id, PresentingWithSymptoms) {
+                        context.modify_isolation_status(person_id, true).unwrap();
```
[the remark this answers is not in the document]
KOVALW: Ya I think ultimately this will be removed by using events (#162) but this works for now

src/parameters.rs
```diff
@@ -175,6 +193,35 @@ fn validate_inputs(parameters: &Params) -> Result<(), IxaError> {
...
+        }
+        if intervention_policy_parameters.isolation_delay_period < 0.0 {
+            return Err(IxaError::IxaError(
+                "The isolation delay period must be non-negative.".to_string(),
+            ));
+        }
+    }
+
+    if let Some(facemask_parameters) = parameters.facemask_parameters {
+        if facemask_parameters.facemask_efficacy < 0.0
```
KOVALW, line 217: Nit: I thought fmt would catch this, but `if !(0.0..=1.0).contains(&facemask_parameters.facemask_efficacy)` is the preferred (see the relative infectiousness example above). This applies to the bounded intervention parameters as well

KOVALW (approved):
LGTM! Much improved, do fix the small formatting thing and then I think it's good to go

confunguido (approved):
I have created issues #184 and #185 related to this PR.

PR #158, opened 2025-06-04
Suggestions for isolation policy

confunguido:


src/parameters.rs
```diff
@@ -61,6 +68,7 @@ pub struct Params {
     pub synth_population_file: PathBuf,
     /// The path to the transmission report file
     pub transmission_report_name: Option<String>,
+    pub isolation_policy_parameters: IsolationParameters,
```
ChiragKumar9, line 71: The only problem with this is that we need to specify the parameters everywhere, even in tests that don't depend on these parameters. I still like this suggestion a lot and would suggest it be merged in, but I think how we specify parameters in tests needs to be thought through more as our model gets more advanced.

src/parameters.rs
```diff
@@ -194,6 +202,7 @@ mod test {
             synth_population_file: PathBuf::from("."),
             transmission_report_name: None,
             settings_properties: HashMap::new(),
+            isolation_policy_parameters: HashMap::new(),
```
ChiragKumar9, line 205: So, for instance, here you cannot have `HashMap::new()` but rather `IsolationParameters { ... }`. You could make `isolation_policy_parameters` take an `Option<IsolationParameters>` and then specify none, but this parameter isn't really optional so that feels weird

ChiragKumar9 (approved):
Love the idea, syntax just needs to be cleaned up/made consistent to compile.

erik-rosenstrom:
I am going to close this pull request. The suggested changes have been incorporated into #154

PR #175, opened 2025-06-13
Query susceptibles once when seeding

ChiragKumar9:


src/infection_propagation_loop.rs
```diff
@@ -57,35 +57,32 @@ fn query_susceptibles_and_seed(
     to_seed: u64,
     seed_fn: impl Fn(&mut Context, PersonId),
 ) -> Result<(), IxaError> {
+    let mut susceptibles = context
+        .query_people((InfectionStatus, InfectionStatusValue::Susceptible))
+        .into_iter();
+    trace!("Found {} susceptibles to seed.", susceptibles.len());
```
KOVALW, line 63: Check if `susceptibles.len() < to_seed` instead of the for loop if statement.

src/infection_propagation_loop.rs
```diff
@@ -57,35 +57,32 @@ fn query_susceptibles_and_seed(
...
-            (InfectionStatus, InfectionStatusValue::Susceptible),
-        );
-        match person {
-            Some(person) => {
-                seed_fn(context, person);
-            }
-            None => {
-                return Err(IxaError::IxaError("The number of people to seed with initial conditions has surpassed the population size.".to_string()));
-            }
+        let person = susceptibles.next();
```
KOVALW, line 65: Using next is going to be a deterministic selection, which we don't necessarily want as people's ID's are going to cluster by properties in some synthetic populations. To resolve this, we could use `let id = context.sample_range(Rng, 0..len(susceptibles))` and `let person = susceptibles.remove(id)` to get a person during the for loop. We could alternatively, if we're willing to `cargo add rand` again, we could use `rand::IteratorRandom::choose_multiple()`.

KOVALW (changes requested):
A couple of things for improving performance and making sure we don't deterministically assign the initial individuals

src/infection_propagation_loop.rs
ChiragKumar9: Such a good point, thank you!

KOVALW (commented):
This LGTM. I do wonder how declaring incidence will interact with small population test cases. Beginning with a single infection is such a common use case, but here it would be I(0)~Bin(n, 1/n), which will have a considerable concentration at zero. This will likely only impact integration-type tests, but if we're fine with starting with either higher initial cases (like ~5-10ish infecteds) or allowing many zeroes to run because they don't take simulation time, then this should be fine.

I don't want to throw errors in the case of I(0)=0. It should be ok to assume that it's possible to start with zero infected individuals, with importation occurring later.

src/infection_propagation_loop.rs
```diff
@@ -54,38 +53,26 @@ fn schedule_recovery(context: &mut Context, person: PersonId) {
...
-                seed_fn(context, person);
-            }
-            None => {
-                return Err(IxaError::IxaError("The number of people to seed with initial conditions has surpassed the population size.".to_string()));
-            }
+) {
+    let susceptibles = context.query_people((InfectionStatus, InfectionStatusValue::Susceptible));
+    for person in susceptibles {
+        // We use a random number to determine whether to seed this person.
+        if context.sample_bool(InfectionRng, proportion_to_seed) {
```
confunguido, line 62: I would prefer if we could just sample without replacement from the population, but I think this works fine for now.

confunguido (approved):

PR #189, opened 2025-06-25
Implement minimal integration tests

eqmooring:
# Background
We need to perform integration tests that will verify that `ixa-epi-isolation` is performing as expected. These tests can take various forms, but one of the core approaches is that we will simulate the ABM using assumptions that correspond to the assumptions embedded in compartmental ODE models and then we will compare the results from these different modeling approaches. This branch is based on code initially developed in the branches https://github.com/CDCgov/ixa-epi-isolation/tree/pgv5_explore_ode_comparison and https://github.com/CDCgov/ixa-epi-isolation/tree/wtk-pgv5-ode-comparison.

# What has changed
Numerous files have been added to perform a minimal set of integration tests. These include a Python-based approach that uses `abmwrappers` to run repeated simulations in Ixa (see `scripts/call_integration_test.py`), R scripts that set up inputs for the Ixa simulations (specifically, synthetic populations and empirical infectiousness rate functions) and that numerically solve the ODE models, an RMarkdown report for "testing" by visual inspection, and a bash script that runs everything. See `tests/integration/README.md` for more details about where different scripts are stored and how to perform the tests.
 
# What are the outstanding issues
There's a need to add (many) more integration tests beyond the basics included here. But in addition to that need, there are some more fundamental issues to discuss/resolve:

- The current approach to implementing these integration tests is fairly manual with hard-coding of parameters in various scripts. There may be ways to automate this process, but this could be challenging and it's not clear that doing so is worth the effort, given that different tests use different combinations of inputs and compare to different ODE output.
- Review the file structure. Currently, there are scripts such as `create_integration_test_pops.R` in the `scripts` directory, but would it be better if more scripts related to integration testing were all kept together in the `tests/integration` directory?
- The approach is currently based on visual inspection of plots. There is scope to make comparisons more formal and quantitative, but we'd have to think about how to integrate this into a workflow.

A few minor things:
- Deal with pre-commit issues. Capitalizing the names of compartments (e.g., S, I, and R) in variable names violates the requirements of the `object_name_linter`; is this a concern?
- The integration testing process requires a few R packages (e.g., `deSolve` for ODEs). How are we handling R package management in this repo?


tests/integration/integration_tests_report.rmd
```diff
@@ -0,0 +1,105 @@
+---
+title: "Ixa-epi-isolation integration tests and documentation"
+output: html_document
+---
+
+# Preface
+
+This report is intended to serve three purposes:
+
+1) To generate figures and other output that, by visual inspection, serve as integration tests that demonstrate that Ixa-epi-isolation is performing as expected. The primary conceptual basis for these tests is to compare output from Ixa with output from ordinary differential equation models.
```
KOVALW, line 10: In general, we use "ixa" and "ixa-epi-isolation", all lowercase

KOVALW (changes requested):
Overall LGTM for a first pass; I added some small points for clarity in the Rmd and in the code. I think you'll want to rebase to main so we can ensure this setup works with the current src directory. For now, I would also follow pre-commit's requests to use snake case, and simply keep variable names like `ixa_sir_df`.

renv is troublesome to work with, at the least, so I would say that for now we stick with manual installs of the four requisite packages and once we move integration tests to a separate space, we can revisit the issue

KOVALW (approved):
I don't see anything that necessarily needs to be changed for a first pass integration test, but I do want to bring something to your attention:

The `query_susceptibles_and_seed` function of the infection propagation loop takes in the initial incidence for the simulations and uses it not in a deterministic sense to be multiplied by the population size and select that number of people from the population to seed, but rather it considers each susceptible in the population to be a Bernoulli random variable at that probability. 

This creates a binomial distribution of initial conditions where $n$ is the population size and $p$ is the initial incidence. In this model, given your parameters, 36.6% of simulations begin with zero infecteds. 

I would play with initial incidence (maybe even just 0.05 rather than 0.01) and population size (to reduce variance of stochastic extinction) to make the results a little more predictable.

tests/integration/integration_tests_report.rmd
```diff
@@ -0,0 +1,505 @@
...
+expected_init <- dbinom(
+  x = inital_infect_vals,
+  size = sim_pop_size, prob = 0.01
+)
+expected_init_df <- data.frame(
+  count_infectious = inital_infect_vals,
+  expected_prop = expected_init
+)
+
+ggplot(ixa_sir_initial_df, aes(x = count_infectious)) +
```
KOVALW, line 228: I think it's ok to exclude this plot - we have checks now to verify the distribution in ixa. If you want to include it still, I would suggest moving it to the parameters background section where you discuss seeding infectious individuals

KOVALW (commented):
I think this material is great to have - especially the discussion of final epidemic size and $R_0$. I think some consistency across figure would be great. Either showing an IQR or credible interval around the mean count of individuals in each compartment over time (for simulations in which $R(t=50)> 50$) would show the variation without each needing to plot each trajectory. However, as I mention below, I like seeing all the trajectories for comparing $S(t)$ and $R(t)$, I would just facet them out and show them for each model type so that it's fewer lines per panel.

The math, FWIW, looks correct to me. I would ask that you validate the structured population comparison plot with $R_0=2.216$ instead of 2.5

I've also made some code adjustments and would recommend that you run
```
git pull main
git rebase main
poetry update abmwrappers
```
So that your tests and implementation are up to date.

tests/integration/README.md
```diff
@@ -0,0 +1,10 @@
+# Running integration tests
+
+Currently, the integration tests are performed by visual inspection of figures that are produced with commentary in an HTML report. To generate the report, run (from the root of the repo) `bash tests/integration/run_integration_tests.sh`.
+
+That bash script will generate inputs (by running `scripts/create_integration_test_pops.R`, `scripts/create_integration_test_rate_fns.R`, and `scripts/create_ode_output.R`) that are saved at `tests/input`. Then, the ABM simulations themselves are run in Ixa and the simulation output are saved in the respective sub-directory in `tests/integration`; for example, for the ABM that uses typical SIR assumptions, the simulation output are stored in `tests/integration/sir/data/simulations`. Finally, the HTML report is generated and can be found at `tests/integration/integration_tests_report.html`.
+
+The following actions are required before running the bash script:
+
+- Ensure the following R packages are installed: `tidyverse`, `deSolve`, `arrow`, and `rmarkdown`.
```
KOVALW, line 9: I needed to install `Rmpfr` as well

PR #193, opened 2025-07-01
Policy module for selecting guidance. Closes #191

### LINKED ISSUE #191: We should be able to turn off isolation guidance.

KOVALW:
Modular approach for handling intervention policies by replacing the `updated_guidance` module with a more extensible `policies` module. It also adds support for specifying a `guidance_policy` in the parameters, enabling future flexibility for policy management and defaults to None in order to turn off policies.

### Modularization and Policy Management:

* **Added `policies` module**: The `policies` module replaces the `updated_guidance` module, providing a more scalable structure for managing intervention policies. It includes an `init` function that initializes policies based on the `guidance_policy` parameter.
* **`GuidancePolicies` enum**: Introduced `GuidancePolicies` with an initial variant `Updated` to represent the updated guidance policy.
* **Added `guidance_policy` to `Params`**: The `Params` struct now includes an optional `guidance_policy` field to specify the policy guidance for interventions. This is integrated into the default implementation.


input/input.json
```diff
@@ -27,6 +27,7 @@
     },
     "facemask_parameters": {
       "facemask_efficacy": 0.2
-    }
+    },
+    "guidance_policy": "Updated"
```
erik-rosenstrom, line 31: When we add the previous guidance, it is going to be defined by a different set of parameters than the updated policy. If we keep the name intervention_policy_parameters, the user would have to switch back and forth adding and removing parameters if doing it by hand. Is it possible to have griddler handle dictionaries, so this isn't an issue at scale? Maybe we should think about this more when doing the previous guidance PR rather than here.

erik-rosenstrom: This will also impact the parameter validation checks for policy parameters.

erik-rosenstrom: I think we need to add a check that if the policy is updated the dictionary contains the right values.

erik-rosenstrom (commented):
This is great! I have just a few comments about managing policy parameters.

confunguido (approved):
Please look into the parameter naming and good to merge

PR #198, opened 2025-07-02
Hospitalizations Implementation

erik-rosenstrom:
This is rough draft of the hospitalization implementation.


input/input.json
```diff
@@ -27,6 +27,21 @@
     },
     "facemask_parameters": {
       "facemask_efficacy": 0.2
+    },
+    "hospitalization_parameters": {
+      "mean_delay_to_hospitalization": 1.0,
+      "mean_duration_of_hospitalization": 10.0,
+      "age_groups": [
```
confunguido, line 34: This seems like you could specify something erroneous: 

```
[
{min: 0, max:4, prob:0.5},
{min: 30, max:6, probs:0.5}
]

```

src/hospital_incidence_report.rs
```diff
@@ -0,0 +1,224 @@
+use crate::{
```
confunguido, line 1: The incidence report is straight forward but it would nicer to have a combined report of person properties that include prevalence or incidence.

src/hospitalizations.rs
```diff
@@ -0,0 +1,508 @@
...
+
+use crate::{
+    define_setting_category,
+    parameters::{ContextParametersExt, HospitalizationParameters},
+    population_loader::Age,
+    settings::{ContextSettingExt, SettingProperties},
+    symptom_progression::{SymptomValue, Symptoms},
+};
+
+define_setting_category!(Hospital);
```
confunguido, line 16: I would suggest to not do this at this time.

src/hospitalizations.rs
```diff
@@ -0,0 +1,508 @@
...
+        max: 120,
+        probability: 0.0
+    }
+);
+
+pub trait ContextHospitalizationExt {
+    fn get_hospital_id(&self) -> usize;
+}
+
+impl ContextHospitalizationExt for Context {
```
confunguido, line 43: Please remove for now.

src/hospitalizations.rs
```diff
@@ -0,0 +1,508 @@
...
+        );
+        Ok(())
+    }
+
+    fn setup_hospitalization_event_sequence(
+        &mut self,
+        hospitalization_parameters: HospitalizationParameters,
+    ) {
+        // Clone hospitalization_parameters for each closure to avoid move errors
+        let hospitalization_parameters_for_symptoms = hospitalization_parameters.clone();
```
confunguido, line 149: Why are you cloning the same structure and calling it something else?

src/hospitalizations.rs
```diff
@@ -0,0 +1,508 @@
...
+        let hospitalization_parameters_for_hospitalized = hospitalization_parameters.clone();
+        // Subscribe to individuals being presymptomatic to plan if/when they enter the hospital
+        self.subscribe_to_event(move |context, event: PersonPropertyChangeEvent<Symptoms>| {
+            let hospitalization_parameters = hospitalization_parameters_for_symptoms.clone();
+            if let Some(SymptomValue::Presymptomatic) = event.current {
+                context
+                    .plan_hospital_arrival(event.person_id, hospitalization_parameters.clone())
```
confunguido, line 156: This sequence seems off. I think you first evaluate risk and then plan for arrival, or simply just change person property to true for hospitalized.

src/hospitalizations.rs
```diff
@@ -0,0 +1,508 @@
...
+            }
+        });
+        // Subscribe to individuals being hospitalized to plan when they leave the hospital
+        self.subscribe_to_event(
+            move |context, event: PersonPropertyChangeEvent<Hospitalized>| {
+                let hospitalization_parameters =
+                    hospitalization_parameters_for_hospitalized.clone();
+                if event.current {
+                    context
+                        .plan_hospital_departure(
```
confunguido, line 167: This could be just a set person property

src/hospitalizations.rs
```diff
@@ -0,0 +1,508 @@
...
+
+#[derive(Debug, Clone, Copy, Deserialize, Serialize, PartialEq)]
+
+pub struct HospitalAgeGroups {
+    pub min: u8,
+    pub max: u8,
+    pub probability: f64,
+}
+
+define_person_property_with_default!(
```
confunguido, line 29: Does this really need a default?

confunguido (changes requested):
Please take a look at my comments and I can review again later.

src/hospitalizations.rs
```diff
@@ -0,0 +1,508 @@
...
+
+    fn evaluate_hospitalization_risk(&mut self, person_id: PersonId) -> bool {
+        // Evaluate the risk of hospitalization using the age group probabilities
+        self.sample_bool(
+            HospitalizationRng,
+            self.get_person_property(person_id, HospitalAgeGroup)
+                .probability,
+        )
+    }
+    fn set_hospital_age_group(&mut self) {
```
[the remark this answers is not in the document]
erik-rosenstrom: Using the derived person property macro is giving me some trouble to get the parameters from defining the age groups into the closure to define the person property. Something like ...

```rust
let hospitalization_parameters = context.get_params().hospitalization_parameters.clone();
    let age_groups = hospitalization_parameters
    .and_then(|params| Some(params.age_groups))
    .unwrap_or_default();

    define_derived_property!(HospitalAgeGroup, HospitalAgeGroups, [Age], |age| {
        for group in &age_groups {
            if (group.min..=120).contains(&age) {
                return group.clone();
            }
        }
        match age {
           ...
        }
```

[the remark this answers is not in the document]
KOVALW: This should be done outside of a function and should rely on calling the global_dependency, such as

```rust
define_derived_property!(
  AgeGroup,
  AgeGroupSelection,
  [Age],
  [Params],
  | age, params | {
    ..
  }
)
```

src/hospital_incidence_report.rs
erik-rosenstrom: This has been addressed using the enum approach suggested by Will.

input/input.json
erik-rosenstrom: This has been addressed using the approach we discussed and that Will also suggested.

src/hospitalizations.rs
```diff
@@ -0,0 +1,508 @@
...
+        person_id: PersonId,
+        hospitalization_parameters: HospitalizationParameters,
+    ) -> Result<(), ixa::IxaError> {
+        // get hospital parameters
+        // evaluate hospitalization risk
+        // plan a delay to enter the hospital
+        if self.evaluate_hospitalization_risk(person_id) {
+            // Get the hospital ID for the person                    // Plan the arrival at the hospital
+            let exp =
+                Exp::new(1.0 / hospitalization_parameters.mean_delay_to_hospitalization).unwrap();
```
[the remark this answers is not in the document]
erik-rosenstrom: Just to clarify, you think the user should be able to toggle between deterministic and stochastic implementations?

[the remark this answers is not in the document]
KOVALW: I think the user should be able to specify the time delay distribution (Gamma, Exp, read from a file, or in the simplest case, fixed). I think for this implementation, we should keep it as fixed (as I show above) and file an issue about adjusting the time delay to account for variability in reporting to the hospital.

src/hospitalizations.rs
erik-rosenstrom: We are removing age group person properties for the time being.

src/contact.rs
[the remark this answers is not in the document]
erik-rosenstrom: I will rebase after your PR and AnySettings are merged into main

src/hospitalizations.rs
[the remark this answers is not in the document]
erik-rosenstrom: Even when I was trying to use GlobalParams, it was still giving me trouble. Guido has suggested dropping the person property for the time being and just using age groups when evaluating risk.

src/hospitalizations.rs
```diff
@@ -0,0 +1,367 @@
...
+            },
+        );
+    }
+
+    fn evaluate_hospitalization_risk(&mut self, person_id: PersonId) -> bool {
+        // Evaluate the risk of hospitalization using the age group probabilities
+
+        let age = self.get_person_property(person_id, Age);
+        let hospitalization_parameters = self.get_params().hospitalization_parameters.clone();
+        if let Some(hospitalization_parameters) = hospitalization_parameters {
```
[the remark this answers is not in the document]
erik-rosenstrom: I removed it and added other checks to ensure the parameterization is valid

.gitignore
```diff
@@ -1,6 +1,10 @@
 #####
 # Exclude many data and output file types by default
 
+#experiment
+scripts/experiments/benchmarking/jul25-assessment/data/*
```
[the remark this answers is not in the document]
erik-rosenstrom: I have switched to .git/info/exclude for now

KOVALW (approved):
LGTM!

erik-rosenstrom:
Going to remove the for loop implementation of age to age group in favor of a map

src/hospitalizations.rs
```diff
@@ -0,0 +1,399 @@
...
+
+trait ContextHospitalizationInternalExt {
+    fn plan_hospital_arrival(&mut self, person_id: PersonId) -> Result<(), ixa::IxaError>;
+    fn plan_hospital_departure(&mut self, person_id: PersonId) -> Result<(), ixa::IxaError>;
+    fn setup_hospitalization_event_sequence(&mut self);
+    fn evaluate_hospitalization_risk(&mut self, person_id: PersonId) -> bool;
+    fn set_age_group_mapping(&mut self);
+}
+
+impl ContextHospitalizationInternalExt for Context {
```
KOVALW, line 70: Use the `PluginContext` syntax as in #222:
```suggestion
trait ContextHospitalizationInternalExt: PluginContext + ContextRadnomExt + ContextPeople Ext{
```
followed by the `impl` block:
```
impl ContextHospitalizationInternalExt for Context {}
```

src/hospitalizations.rs
```diff
@@ -0,0 +1,399 @@
...
+    pub probability: f64,
+}
+
+#[derive(Default)]
+struct HospitalDataContainer {
+    age_group_mapping: HashMap<u8, HospitalAgeGroups>,
+}
+
+impl HospitalDataContainer {
+    fn set_age_group_mapping(&mut self, age_groups: &[HospitalAgeGroups]) {
```
KOVALW, line 32: I think this is a little clearer

```rust
for i in 1..(age_groups.len()) {
  let grp = &age_groups[i];
  let max = if i == age_groups.len() - 1 {
    121
  } else {
    &age_groups[i+1].min
  };
  for age in grp.min..max {
    self.age_group_mapping.insert(age, *grp)
  }
}
```

src/hospitalizations.rs
```diff
@@ -0,0 +1,508 @@
...
+                alpha: 0.0,
+                itinerary_specification: None,
+            },
+        )
+        .unwrap();
+    let hospitalization_parameters = context.get_params().hospitalization_parameters.clone();
+    if let Some(hospitalization_parameters) = hospitalization_parameters {
+        context.set_hospital_age_group();
+        context.setup_hospitalization_event_sequence(hospitalization_parameters.clone());
+    } else {
```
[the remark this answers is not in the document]
confunguido: Thanks Erik. Should it be an option or do we want to always have some hospitalization parameters?

src/hospital_incidence_report.rs
```diff
@@ -0,0 +1,258 @@
...
+    });
+}
+
+fn create_hospital_incidence_report(
+    context: &mut Context,
+    file_name: &str,
+    report_type: HospitalizationReportType,
+) -> Result<(), IxaError> {
+    match report_type {
+        HospitalizationReportType::Incidence => {
```
confunguido, line 42: Since this doesn't address my comment, I think you should just let this module about incidence (since that's the name of the module). Don't provide an option for prevalence, as Will said, that's just part of the periodic report.

confunguido (changes requested):
Please address the comments before merging. I will create an issue for the unified incidence reports for person properties, but for now, don't have an option for prevalence.

src/hospitalizations.rs
```diff
@@ -0,0 +1,389 @@
...
+            },
+        );
+        Ok(())
+    }
+    fn plan_hospital_departure(&mut self, person_id: PersonId) -> Result<(), ixa::IxaError> {
+        // get hospital parameters
+        // select a duration of hospitalization
+        // plan to leave the hospital after the duration
+        trace!("Planning hospital departure for person {person_id}");
+        let mean_duration_of_hospitalization = self
```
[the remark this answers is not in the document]
KOVALW: This was my suggestion. Erik originally had an Exponential. I think it doesn't make sense to assume a random distribution in a hard-coded way for this first pass and suggested a fixed time delay, which would be rpelaced by user-specified distributions later.

src/hospital_incidence_report.rs
```diff
@@ -0,0 +1,259 @@
...
+                            context,
+                            event.person_id,
+                            age,
+                            "Arrival".to_string(),
+                        );
+                    }
+                },
+            );
+        }
+        HospitalizationReportType::Prevalence => {
```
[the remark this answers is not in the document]
erik-rosenstrom: This makes sense.

src/hospitalizations.rs
```diff
@@ -0,0 +1,389 @@
...
+    let initialization_check = hospitalization_parameters
+        .age_groups
+        .iter()
+        .any(|grp| grp.probability > 0.0);
+    if initialization_check {
+        context
+            .store_transmission_modifier_values(
+                InfectionStatusValue::Infectious,
+                Hospitalized,
+                &[(true, 0.0)],
```
confunguido, line 154: I don't think we had agreed on this. I thought we were only using the person property.

src/hospitalizations.rs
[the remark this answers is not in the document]
erik-rosenstrom: I have added the exponential distribution back for now

src/hospitalizations.rs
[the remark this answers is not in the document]
erik-rosenstrom: This implementation enables us to get by without a Boolean that we would need to toggle. The defaults parameters for tests enable us to have to add the parameters to test.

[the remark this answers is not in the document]
confunguido: What would happen if you initialize the module with probability 0.0?

confunguido (approved):

PR #201, opened 2025-07-03
Replaced algorithm in `get_contact_internal` with a much faster algorithm.

RobertJacobsonCDC:
The algorithm runs in far fewer steps with very high probability. 

This algorithm also calls [`filter_people`](https://github.com/CDCgov/ixa/pull/437) instead of calling `match_person` in a loop. The `filter_people` call runs in 1/3rd the time as calling `match_person` in a loop. But `filter_people` hasn't been merged yet, and so hasn't been released. You'll want to adjust this PR if you merge it before it is.

The model runs in ~5 seconds with this algorithm compared to ~60 seconds with the original algorithm.

page1
page2

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -348,30 +348,53 @@ impl ContextSettingInternalExt for Context {
     ) -> Result<Option<PersonId>, IxaError> {
         let members = self.get_setting_members_internal(setting_type, setting_id);
         if let Some(members) = members {
-            if !members.contains(&person_id) {
+            if members.len() == 1 {
+                // Only contains `person_id`; no one else to infect.
+                return Ok(None);
+            }
+
+            // Instead of scanning the entire set of contacts, collecting those that match the
+            // query, and then randomly selecting, we randomly select and then check the selected
+            // person to see if they match the query and repeat if they do not. We set a max number
+            // of guess-and-checks to be some number n. One can write down a closed form formula
+            // for the expected number of steps this strategy takes. The original strategy takes
+            // c + 1 steps (c checks and 1 random selection), where c is the number of contacts
+            // (excluding infector). If a is the percentage of contacts that match the query (a for
+            // "alive"), it can be shown that the "break even" point,
+            //    E[runtime of this strategy] = c + 1,
+            // occurs when a = 2/(c + 3). Notice this is independent of n. The number of
+            // guess-and-checks needed to find a match to the query is geometrically distributed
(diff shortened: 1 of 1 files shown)


src/settings.rs
```diff
@@ -348,30 +348,53 @@ impl ContextSettingInternalExt for Context {
...
+            }
+
+            if !q.get_query().is_empty() {
                 // If the query is not empty, we match setting members to the query
-                for contact in member_iter {
-                    if self.match_person(*contact, q) {
-                        contacts.push(*contact);
-                    }
-                }
+                self.filter_people(&mut contacts, q);
```
RobertJacobsonCDC, line 397: This is the new method in ixa. You'll need to use this until it lands:

```rust
contacts.retain(|contact| self.match_person(*contact, q));
```

src/settings.rs
```diff
@@ -348,30 +348,53 @@ impl ContextSettingInternalExt for Context {
...
+            // c + 1 steps (c checks and 1 random selection), where c is the number of contacts
+            // (excluding infector). If a is the percentage of contacts that match the query (a for
+            // "alive"), it can be shown that the "break even" point,
+            //    E[runtime of this strategy] = c + 1,
+            // occurs when a = 2/(c + 3). Notice this is independent of n. The number of
+            // guess-and-checks needed to find a match to the query is geometrically distributed
+            // and has expected value 1/a. Note this is independent of c. At the break even point,
+            // the expected number of guess-and-checks needed is
+            //     1/a = (c + 3)/2.
+            // We set n = (c + 3)/2 as a heuristic.
```
RobertJacobsonCDC, line 370: A question for the statisticians: Is there a way to formulate an "optimal" choice of $n$ given that we don't know $a$? I couldn't think of a way.

ChiragKumar9:
From the peanut gallery, just reading @RobertJacobsonCDC’s math, why not do step 2 but remove people that are rejected from the set? Then the number of samples to draw is at most the size of the set.

It would make more sense for collections of people to be stored as hash sets for easy removal, so maybe this prompts a broader question about whether the return from a query should be a hash set (or index set) rather than a vector.

RobertJacobsonCDC:
> ...why not do step 2 but remove people that are rejected from the set?

Good question! The answer is, because I don't want to create a copy of the whole vector if I don't have to, and so I don't want to mutate it. It's also the reason I don't remove `person_id` ("the infector").

RobertJacobsonCDC:
Closed in favor of #205 for git hygiene reasons.

PR #203, opened 2025-07-08
Use `sample_people()` method

KOVALW:
### Changes
* Updated the `ixa` dependency from version `0.1.1` to `0.2.0` in `Cargo.toml` to incorporate the latest features and fixes.
* Replaced the use of `context.query_people` with `context.sample_people`, which directly samples a subset of susceptibles based on the calculated number (`n`)
    - Previous implementation had the proportion being calculated on susceptibles remaining, meaning that seeding infections impacted the proportion of "recovered" in the population. Now `n` is strictly based on population size

### DIFF
--- a/src/infection_propagation_loop.rs
+++ b/src/infection_propagation_loop.rs
@@ -1,4 +1,5 @@
 use core::f64;
+use statrs::distribution::Binomial;
 
 use crate::infectiousness_manager::{
     evaluate_forecast, get_forecast, infection_attempt, Forecast, InfectionContextExt,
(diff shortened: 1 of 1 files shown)


src/infection_propagation_loop.rs
```diff
@@ -49,19 +48,25 @@ fn schedule_recovery(context: &mut Context, person: PersonId) {
...
 ) {
-    let susceptibles = context.query_people((InfectionStatus, InfectionStatusValue::Susceptible));
+    #[allow(clippy::cast_precision_loss)]
+    let n = proportion_to_seed * context.get_current_population() as f64;
+
+    #[allow(clippy::cast_possible_truncation, clippy::cast_sign_loss)]
+    let susceptibles = context.sample_people(
+        InfectionRng,
+        (InfectionStatus, InfectionStatusValue::Susceptible),
+        n as usize,
```
confunguido, line 66: What happens if there aren't enough susceptibles?

PR #205, opened 2025-07-08
Replace algorithm in `get_contact_internal` with a much faster algorithm. Fixes #195.

### LINKED ISSUE #195: Querying in get_contact_internal
To get a contact, we are requesting a query. Currently this query is `Alive`. However, querying inside settings is not implemented in ixa. The current implementation of this query is very slow as it goes through the entire population of the setting and runs `match_person`. This isn't a problem for small settings but as the population gets larger, this query becomes more of a problem for census tracts. For instance, in a population of 100,000 people, the run time went from 54 seconds to 5 seconds when removing this query. 

```rust
if q.get_query().is_empty() {
        // If the query is empty we push members directly to the vector
        for contact in member_iter {
            contacts.push(*contact);
        }
    } else {
        for contact in member_iter {
            if self.match_person(*contact, q) {
                contacts.push(*contact);
            }
        }
    }

```

RobertJacobsonCDC:
This PR addresses performance issues in `settings::ContextSettingInternalExt::get_contact_internal()` when the size of the vector of candidate contacts is large.

- We avoid copying the large vector when it isn't necessary.
- We use a bounded guess-and-check strategy for finding a contact that satisfies the given query. 
- We bound the guess-and-check attempts using a reasonable heuristic. 
- If we hit the bound, we fall back to the original "filter first, then select" that always succeeds if there is a valid contact. This slow path occurs only with low probability.

A first draft of this PR with mathematical details is here: https://github.com/CDCgov/ixa-epi-isolation/pull/201.

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -615,32 +615,56 @@ pub trait ContextSettingExt:
         setting: &dyn AnySettingId,
         q: T,
     ) -> Result<Option<PersonId>, IxaError> {
-        let members = self.get_setting_members(setting);
-        if let Some(members) = members {
-            if !members.contains(&person_id) {
+        if let Some(members) = self.get_setting_members(setting) {
+            if members.len() == 1 {
+                // Only contains `person_id`; no one else to infect.
+                return Ok(None);
+            }
+            // Instead of scanning the entire set of contacts, collecting those that match the
+            // query, and then randomly selecting, we randomly select and then check the selected
+            // person to see if they match the query and repeat if they do not. We set a max number
+            // of guess-and-checks to be some number n. One can write down a closed form formula
+            // for the expected number of steps this strategy takes. The original strategy takes
+            // c + 1 steps (c checks and 1 random selection), where c is the number of contacts
+            // (excluding infector). If a is the percentage of contacts that match the query (a for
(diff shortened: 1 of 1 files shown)


RobertJacobsonCDC:
Fixed break even computation: $a = \frac{2}{c + 1}$

```
Wolfram Language 14.0.0 Engine for Mac OS X ARM (64-bit)
Copyright 1988-2023 Wolfram Research, Inc.

In[1]:= f[a_, c_, n_]:= Sum[2*k*a*(1-a)^(k-1), {k, 1, n}] + (2*n + c + 1)*(1-a)^n

In[2]:= f[a, c, n]

                                                n          n
               n                 2 (-1 + (1 - a)  + (1 - a)  a n)
Out[2]= (1 - a)  (1 + c + 2 n) - --------------------------------
                                                a

In[3]:= Solve[f[a, c, n] == c + 1, a, Reals]

                                       2
Out[3]= {{a -> ConditionalExpression[-----, c > 1 || c < -1]}}
                                     1 + c
```

confunguido:
Closing in favor of https://github.com/CDCgov/ixa-epi-isolation/pull/234

PR #208, opened 2025-07-09
Experiment class benchmarking script

erik-rosenstrom:
This PR introduces a small script to benchmark simulations with different simulation population sizes. It uses the experiment class developed as part of the abmwrappers package. It records the average run time, size of the synthetic population, and attack rate.


scripts/experiments/benchmarking/jul25-assessment/input/config.yaml
```diff
@@ -8,7 +8,7 @@ experiment_conditions:
 local_path:
   default_params_file: /home/xui6/github/ixa-epi-isolation/scripts/experiments/benchmarking/jul25-assessment/input/base.json
   griddle_file: /home/xui6/github/ixa-epi-isolation/scripts/experiments/benchmarking/jul25-assessment/input/griddle.json
-  exe_file: target/release/epi-isolation
+  exe_file: ./target/release/epi-isolation
```
KOVALW, line 11: Was now relative path causing issues otherwise?

KOVALW (changes requested):
Some things to change to use the more ergonomic functions offered by the package. From this implementation it's clear that the info available to the user isn't well-communicated and there's some info that would be better to have produced automatically for the user. I'm going to address some of these issues before this PR should be merged in

scripts/experiments/benchmarking/jul25-assessment/input/config.yaml
erik-rosenstrom: Yeah I was having some issues with the paths for the target and experiment definition. I think the issue was that my terminal was not cd'ed into the correct folder when I was running the script.

KOVALW: Good to know - I'll look into this as well and see if we want some handling for these cases or if we want to push to an ETL PR on the load_config method

erik-rosenstrom: more specifically, when I run the script in my ide, a new python terminal is opened from root of WSL (home/xui6/) which is a little sneaky.

RobertJacobsonCDC:
The [execution stats feature](https://github.com/CDCgov/ixa/pull/427) we just merged into ixa core, and the `profiling` module ([PR on epi-isolation](https://github.com/CDCgov/ixa-epi-isolation/pull/209)) are intended for human-friendly feedback at development time as opposed to use in a benchmarking suite like this one, BUT I wonder if it would be worth breaking out finer-grained execution time statistics that could be used in something like this. I'm thinking in particular of:

- Separating out the time to load data / initialize the Context from the time to execute the simulation after initialization.
- Capturing peak memory use (as in [the execution stats feature](https://github.com/CDCgov/ixa/pull/427)).

Also consider using CPU time instead of wall time (or both).

KOVALW:
File management plan/docs needed. The "scripts" folder is getting bloated. Include viz of output when marking ready for review

erik-rosenstrom:
image
image

erik-rosenstrom:
image

KOVALW (approved):
LGTM. Could be nice to include a step in the benchmark script to assert that the population files exist and include a `benchmark` as follows in the Makefile:

```
benchmark: experiments/examples/benchmarking/scripts/benchmark.py $(BIN)
  poetry run python $(<)
```

so we can just call `make benchmark`. I'd also suggest storing the output figures in a directory called `/figures/` instead of `/data/`

confunguido (approved):

PR #209, opened 2025-07-11
Created the `profiling` module for plan per second stats.

RobertJacobsonCDC:
Created the `profiling` module that provides simple mechanism to count events during a simulation and report on total accumulated counts and per-second rates.
Added plan per second statistics using the new `profiling` module.

# The module-level documentation

A simple mechanism to count events during a simulation and report on total simple
mechanism to count events during a simulation and report on total accumulated counts
and per-second rates.

```
Event                   Count  Rate (per sec)
---------------------------------------------
property progression   12,888         2428.00
recovery                9,091         1712.67
accepted infection      8,988         1693.27
forecasted infection   27,171         5118.81

Infection Forecasting Efficiency: 33.08% (8,988 accepted of 27,171 forecasted)
```

This module provides an interface for collecting statistics on how frequently
certain events occur during a simulation. It is designed to track both the total
count of events and the event rate (count per second) over time.

Although originally intended for measuring how often plans are processed,
the mechanism is general-purpose and can be used to track any discrete event
in the simulation, for example:

- Monitoring the frequency of specific agent behaviors
- Tracking usage patterns of system resources
- Measuring throughput of scheduling or processing queues

The mechanism is very simple:

- **Count accumulation**: Keep track of how many times an event has occurred by calling
  `context.increment_named_count()` with the name of the event (a `&'static str`).
- **Rate estimation**: The first time `context.increment_named_count()` with **any** name
  the global start time is recorded. Display the count and computed rate (e.g., plans per
  second) with `context.print_named_counts()`.

The names `"forecasted infection"` and `"accepted infection"` are treated specially by
`print_named_counts()` in that the forecast efficiency, defined as

```ignore
accepted_forecast_count / forecast_count * 100.0
```

is also computed and printed.

# Usage

To use this module, increment the counter in your simulation loop, plan, or event handler.
Here is an example with

```rust
context.add_plan(next_time, move |context| {

    // Increment the count for the "forecasted infection" event
    context.increment_named_count("forecasted infection");

    if evaluate_forecast(context, person, forecasted_total_infectiousness) {
        if let Some(setting_id) = context.get_setting_for_contact(person) {
            let str_setting = setting_id.setting_type.get_name();
            let id = setting_id.id;
            if let Some(next_contact) =
                infection_attempt(context, person, setting_id)
            {
                // Increment the count for "accepted infection"
                context.increment_named_count("accepted infection");

                context.infect_person(next_contact, Some(person), Some(str_setting), Some(id));
            }
        }
    }
    // Continue scheduling forecasts until the person recovers.
    schedule_next_forecasted_infection(context, person);
});
```

### DIFF
--- a/src/infection_propagation_loop.rs
+++ b/src/infection_propagation_loop.rs
@@ -5,6 +5,7 @@ use crate::infectiousness_manager::{
     InfectionData, InfectionDataValue, InfectionStatus, InfectionStatusValue,
 };
 use crate::parameters::{ContextParametersExt, Params};
+use crate::profiling::ContextProfilingExt;
 use crate::rate_fns::{load_rate_fns, InfectiousnessRateExt};
 use crate::settings::ContextSettingExt;
 use ixa::{
@@ -21,6 +22,7 @@ fn schedule_next_forecasted_infection(context: &mut Context, person: PersonId) {
     }) = get_forecast(context, person)
     {
         context.add_plan(next_time, move |context| {
+            context.increment_named_count("forecasted infection");
             if evaluate_forecast(context, person, forecasted_total_infectiousness) {
                 if let Some(setting_id) = context.get_setting_for_contact(person) {
                     let str_setting = setting_id.setting_type.get_name();
@@ -28,6 +30,7 @@ fn schedule_next_forecasted_infection(context: &mut Context, person: PersonId) {
                     if let Some(next_contact) =
                         infection_attempt(context, person, setting_id)
                     {
+                        context.increment_named_count("accepted infection");
                         trace!("Person {person}: Forecast accepted, setting type {str_setting} {id}, infecting {next_contact}");
                         context.infect_person(next_contact, Some(person), Some(str_setting), Some(id));
                     }
@@ -43,6 +46,7 @@ fn schedule_recovery(context: &mut Context, person: PersonId) {
     let infection_duration = context.get_person_rate_fn(person).infection_duration();
     let recovery_time = context.get_current_time() + infection_duration;
     context.add_plan(recovery_time, move |context| {
+        context.increment_named_count("recovery");
         trace!("Person {person} has recovered at {recovery_time}");
         context.recover_person(person);
     });
@@ -132,6 +136,7 @@ mod test {
         distribution::{ContinuousCDF, Discrete, Poisson, Uniform},
     };
 
+    use crate::profiling::ContextProfilingExt;
     use crate::{
         define_setting_type,
         infection_propagation_loop::{
@@ -676,6 +681,7 @@ mod test {
                     match event.current {
                         Masking::None => {
                             context.add_plan(t + nonmasking_duration, move |context| {
+                                context.increment_named_count("mask wearing");
                                 context.set_person_property(
                                     event.person_id,
                                     MaskingStatus,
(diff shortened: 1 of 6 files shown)


confunguido:
This looks good. In terms of usage, is it always enabled or do you have to pass a flag to activate it?

RobertJacobsonCDC:
This is failing in CI because it needs to be stacked on top of #231. (In other words, I screwed up my branches somehow and need to push fixes.)

KOVALW (changes requested):
Straightforward and already seems useful after running a few parameter sets to inspect (the itinerary modification effect on runtime varies widely while scheduling forecasts takes up a more consistent amount of runtime). It would be helpful to run a parameter sweep and store the results associated with each key value (number of events, duration, total runtime) in a CSV. The parameters can be added in post, so they don't need to be printed to the csv

RobertJacobsonCDC:
I placed counters and spans in a few places mostly as a demonstration, so you will probably want to change those, or changes their names, or something. Maybe once we approve the PR one of you can add a commit to this PR with those minor changes before merging to main.

src/main.rs
```diff
@@ -68,6 +71,7 @@ fn main() {
         hospital_incidence_report::init(context)?;
 
         Ok(())
-    })
-    .unwrap();
```
RobertJacobsonCDC, line 72: We still want to expose this error.

src/profiling.rs
```diff
@@ -0,0 +1,671 @@
...
+            ]);
+        }
+
+        println!();
+        print_formatted_table(&rows);
+    }
+
+    fn print_named_spans(&self) {
+        if self.open_span_count != 0 {
+            println!("OPEN SPAN COUNT NONZERO: {}", self.open_span_count);
```
RobertJacobsonCDC, line 277: Remove these two print debugging artifacts. The new mechanism should rule these cases out (except in a `Test` context where we intentionally ignore these cases).

src/natural_history_parameter_manager.rs
```diff
@@ -200,7 +200,7 @@ mod test {
                 assert_eq!(msg, "An assignment function for this parameter has already been registered.".to_string());
             }
             Some(ue) => panic!(
-                "Expected an error that an assignment function for this parameter has already been registered. Instead got {:?}",
+                "Expected an error that an assignment function for this parameter has already been registered.. Instead got {:?}",
```
KOVALW, line 203:
```suggestion
                "Expected an error that an assignment function for this parameter has already been registered. Instead got {:?}",
```

src/infection_propagation_loop.rs
```diff
@@ -22,7 +23,10 @@ fn schedule_next_forecasted_infection(context: &mut Context, person: PersonId) {
     }) = get_forecast(context, person)
     {
         let infection_plan = context.add_plan(next_time, move |context| {
+            let _span = open_span("evaluate and schedule next forecast");
+            increment_named_count("forecasted infection");
             if evaluate_forecast(context, person, forecasted_total_infectiousness) {
+                increment_named_count("accepted infection");
```
KOVALW, line 29: This is to accept an infection attempt, not the infection itself

KOVALW (commented):
The implementation LGTM, I've made some small comments on the named counts that we track. 

One thought I keep having is that the increment named counts either serve as a dummy count for the span they're actually tracking such as the `infection_attempt` being equal to `accepted infection` and `schedule_next_forecasted_infection` being equal to `forecasted infection`, or could be spans instead, such as `recovery` just becoming `schedule_recovery`. The data manipulations on counts could be done in the same way, so I'm wondering what the advantage is of just having the count and average rate over time instead of true simulation time.

I'm happy to be convinced that the incremented count events are useful, but it just seems like extra report/tracking effort without information gain.

PR #211, opened 2025-07-11
First test for failure on modified itinerariese and forecast

KOVALW:
This test evaluates forecasts during  infector itinerary modification and includes a new test case to show how evaluation fails if modified itinerary occurs between forecast and evaluation.


confunguido (approved):
The test looks fine but I would let it break so we can fix it when we fix the bug.

PR #217, opened 2025-07-16
Correct self modifier test using events in schedule_forecast

KOVALW:
Addition of itinerary change event-driven forecast updates to ensure forecasted infectiousness is always above current infectiousness.

### Enhancements to itinerary-based infectiousness management:
* **Event-driven itinerary updates**: Added `ItineraryChangeEvent` to emit events when a person's itinerary changes, enabling dynamic adjustments to infection plans based on itinerary modifications.
* **Infectiousness multiplier tracking**: Updated `modify_itinerary` to calculate and emit the previous and current infectiousness multipliers for individuals during itinerary changes.
* **Dynamic infection plan cancellation**: Subscribed to `ItineraryChangeEvent` in `schedule_next_forecasted_infection` to cancel and reschedule infection plans dynamically when itinerary changes increase infectiousness. 
* **Moved test of itinerary modification impacts**: Moved `test_forecast_during_self_itinerary_modification` to validate infection forecasts during itinerary changes that affect infectiousness.

### DIFF
--- a/src/infection_propagation_loop.rs
+++ b/src/infection_propagation_loop.rs
@@ -6,7 +6,7 @@ use crate::infectiousness_manager::{
 };
 use crate::parameters::{ContextParametersExt, Params};
 use crate::rate_fns::{load_rate_fns, InfectiousnessRateExt};
-use crate::settings::ContextSettingExt;
+use crate::settings::{ContextSettingExt, ItineraryChangeEvent};
 use ixa::{
     define_rng, trace, Context, ContextPeopleExt, ContextRandomExt, IxaError, PersonId,
     PersonPropertyChangeEvent,
@@ -20,7 +20,7 @@ fn schedule_next_forecasted_infection(context: &mut Context, person: PersonId) {
         forecasted_total_infectiousness,
     }) = get_forecast(context, person)
     {
-        context.add_plan(next_time, move |context| {
+        let infection_plan = context.add_plan(next_time, move |context| {
             if evaluate_forecast(context, person, forecasted_total_infectiousness) {
                 if let Some(setting_id) = context.get_setting_for_contact(person) {
                     let str_setting = setting_id.setting_type.get_name();
@@ -35,6 +35,12 @@ fn schedule_next_forecasted_infection(context: &mut Context, person: PersonId) {
             }
             // Continue scheduling forecasts until the person recovers.
             schedule_next_forecasted_infection(context, person);
+        });
+        context.subscribe_to_event::<ItineraryChangeEvent>(move |context, event| {
+            if event.person_id == person && event.previous_multiplier < event.current_multiplier {
+                context.cancel_plan(&infection_plan);
+                schedule_next_forecasted_infection(context, person);
+            }
         });
     }
 }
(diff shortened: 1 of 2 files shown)


src/infection_propagation_loop.rs
```diff
@@ -36,6 +36,19 @@ fn schedule_next_forecasted_infection(context: &mut Context, person: PersonId) {
             // Continue scheduling forecasts until the person recovers.
             schedule_next_forecasted_infection(context, person);
         });
+        context.subscribe_to_event::<ItineraryChangeEvent>(move |context, event| {
```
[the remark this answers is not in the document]
erik-rosenstrom: I have gotten around plan cancelations like this in the past by storing plans in a hashmap (in this case in a data container) and having a separate function that would cancel the plan from the event subscription.

src/infection_propagation_loop.rs
```diff
@@ -36,6 +36,19 @@ fn schedule_next_forecasted_infection(context: &mut Context, person: PersonId) {
...
+            // Check if the forecast next_time is still relevant
+            if context.get_current_time() < next_time {
+                // Check if person's infectiousness multiplier is altered without being handled by rejection sampling
+                if (event.person_id == person
```
erik-rosenstrom, line 43: I think it itll help us get around this check too potentially, which seems like it could be done alot for large populations

erik-rosenstrom (changes requested):
My main comments are about code organization regarding the event subscription.

src/infection_propagation_loop.rs
```diff
@@ -45,9 +47,91 @@ fn schedule_recovery(context: &mut Context, person: PersonId) {
...
+                    .keys()
+                    .filter(|&infector| {
+                        (context.is_contact(event.person_id, *infector)
+                            && event.increases_membership)
+                            || (event.person_id == *infector
+                                && event.previous_multiplier < event.current_multiplier)
+                    })
+                    .copied()
+                    .collect();
+
```
KOVALW, line 120: Here is where we should asses `if event.increases_settings` you should get all contacts of the `event.person_id` and query for `Infectious` contacts. Or initialize all infectious individuals into the data container with a `None` plan

KOVALW: This will be addressed in #220

src/infection_propagation_loop.rs
```diff
@@ -45,9 +47,91 @@ fn schedule_recovery(context: &mut Context, person: PersonId) {
...
+        }
+    }
+    fn cancel_forecast(&mut self, person_id: PersonId) {
+        let container = self.get_data_container(ForecastDataPlugin).unwrap();
+        if let Some(active_plan) = container.get_plan(person_id) {
+            self.cancel_plan(&active_plan.clone());
+        }
+    }
+    fn subscribe_to_itinerary_change(&mut self) {
+        self.subscribe_to_event::<ItineraryChangeEvent>(move |context, event| {
```
erik-rosenstrom, line 106: If the person is recovered and they change their itinerary, would infectiousness for active infectors need to be reevaluated? If no it might be worth adding an if statement checking if the person modifying their status is not recovered.

KOVALW:
The stochastic inputs for tests could be interesting, I don't know how that would square with our Defaults but I think it's at least valuable to consider it because you're 100% right that bug-free performance can be a product of parameters. Maybe we could run the basic example with a range of parameters?

PR #221, opened 2025-07-21
Skip seed function if recovered or infection proportion is zero

KOVALW:
Skip the initialization of changing status of susceptible individuals if the proportion to seed is zero. Was causing a sample error for a binomial with mean 0 otherwise. Added a trace statement when the number of individuals to seed is drawn for the population for each call of `query_susceptibles_and_seed`


erik-rosenstrom (approved):
Looks good to me !

PR #223, opened 2025-07-21
Don't call `sample_people` if k is not greater than 0

KOVALW:
`sample_people` panics when sampling 0 people, so it should be skipped in the `query_susceptible_and_seed()` function if we randomly select 0 susceptibles to seed on the binomial draw


confunguido (approved):
This looks fine to me

PR #224, opened 2025-07-23
Remove unnecessary plan cancellation

KOVALW:
The plans that need to be cancelled are handled by the event subscription. This will only cancel already-executed or already-cancelled plans.


erik-rosenstrom (approved):
This change makes sense to me.

PR #234, opened 2025-07-31
Replace get_contact with sample_from_setting_with exclusion. Fixes #195

### LINKED ISSUE #195: Querying in get_contact_internal

confunguido:
The function `get_contact` had several places that slow down the model. 

1. Querying. This isn't necessary at least at this point in our model. If we are going to use queries, we should also find a way to index people. 
2. Checking membership. The way we are using this function, it is a requirement for the person to come be part of the members vector. Using `contains` takes a very long time to run in big settings. In this function `sample_from_setting_with_exclusion` a person is not required to be a member of the setting. 
3. Copying and iterating before sampling. Instead of doing that, I am proposing to sample a person from the vector and checking if it's the same as person id. 

At this point, the model runs 1M people in 1 minute. It can be further improved once we address the periodic reports. 

```
━━━━ Execution Summary ━━━━
Max memory usage:        618.3 MiB
CPU time:                1m 16s 740ms
Wall time:               1m 16s 757ms 123us 771ns
Population:              1000092
Memory per person:       648 B
CPU time per person:     76us 733ns
Wall time per person:    76us 750ns
```

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -1536,7 +1536,7 @@ mod test {
         let setting_id = context.sample_setting(person_a).unwrap();
         let members = context.get_setting_members(setting_id).unwrap();
         assert!(members.contains(&person_a));
-        
+
         assert_eq!(
             Some(person_b),
             context


src/settings.rs
```diff
@@ -1612,173 +1562,8 @@ mod test {
         }
     }
 
-    #[test]
-    fn test_draw_contact_from_transmitter_itinerary() {
```
RobertJacobsonCDC, line 1616: Don't we still want some version of this test?

src/settings.rs
```diff
@@ -609,74 +609,35 @@ pub trait ContextSettingExt:
...
-        let members = self.get_setting_members(setting);
-        if let Some(members) = members {
-            if !members.contains(&person_id) {
-                return Err(IxaError::from(
-                    "Attempting contact outside of group membership",
-                ));
-            }
-            // The setting has one person in it -- this person
+        if let Some(members) = self.get_setting_members(setting) {
             if members.len() == 1 {
```
RobertJacobsonCDC, line 618: You should probably add an `else if members.is_empty() { Ok(None) }` branch. As far as I can tell there isn't a check for this elsewhere in the code, so you're relying on client code to do the right thing. Burry fewer landmines, lose fewer limbs.

RobertJacobsonCDC (changes requested):
The best PRs are the ones that remove a ton of code!

PR #235, opened 2025-07-31
Improve speed of synth population script

confunguido:
Main change is to sample in batches instead of house by house.

### DIFF
--- a/scripts/create_synthetic_population.R
+++ b/scripts/create_synthetic_population.R
@@ -1,4 +1,3 @@
-
 ## =================================#
 ## Setup ---------------
 ## =================================#
@@ -75,7 +74,7 @@ synth_school_df <- as_tibble(pumas_st) |>
   dplyr::mutate(
     lat = as.numeric(INTPTLAT20),
     lon = as.numeric(INTPTLON20)
-  )|>
+  ) |>
   dplyr::select(school_id, lat, lon) |>
   mutate(enrolled = 0)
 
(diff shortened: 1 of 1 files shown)


scripts/create_synthetic_population.R
```diff
@@ -28,98 +31,97 @@ person_variables <- c(
...
-house_counter <- 0
-while (nrow(synth_pop_df) < population_size) {
-  house_counter <- house_counter + 1
+start_time <- Sys.time()
+synth_pop_list <- vector("list", population_size)
+house_counter <- 1
+list_counter <- 1
+sampled_pop <- 0
+while (sampled_pop < population_size) {
+  batch_size <- min(100, population_size - sampled_pop)
```
KOVALW, line 120: 100 is high for small populations and low for large populations as a minimum. The population size difference as batch size will also overestimate the number of houses remaining. We can instead dynamically update based on the current average household size using:
```R
while (sampled_pop < population_size) {
  if (sampled_pop > 0) {
    houses_remaining <- (population_size - sample_pop) / (sampled_pop / house_counter)
    preferred_batch_size <- max(1, floor(0.95 * houses_remaining))
    batch_size <- min(2000, preferred_batch_size)
  } else {
    batch_size <- 30
  }
```
I get performance times around 22s for 1million people instead of 1min 30s without the scaling

KOVALW (commented):
LGTM - I would recommend using my dynamic batch size selection instead of fixing at 100 houses, which will dramatically oversample in smaller populations

KOVALW (approved):
LGTM

PR #238, opened 2025-08-01
better scale parameter for empirical scale

erik-rosenstrom:
Attempting to investigate observed slow down associated with using empirical rate functions. It appears to have turned out to be scale of 1 was generating a high R0 value. It is has been set to 0.05 which generates about ~70% of the population becoming infected in a population of 10,000.

image


KOVALW (approved):
LGTM!

PR #244, opened 2025-08-07
Output profiling data to JSON

RobertJacobsonCDC:
Refactored profiling module.
Added function to write profiling data to JSON.
Added method on `Context` to write profiling data to JSON.

To use this feature:

```rust
  // Somewhere after the simulation finishes, write the profiling data and context's execution statistics to a JSON file.
  context.write_profiling_data();
```

The method uses the path stored in the `profiling_data_path` field of `GlobalParams`.

This PR depends on an unreleased feature in ixa core that publicly exposes `ExecutionStatistics`, so it can't be merged until it lands.

Some open questions:

1. What default behavior do we want? I have a `profiling_data_path` field in `GlobalParams`, but we could have more configuration there, or just not write a file if `profiling_data_path` is `None`, or something else...
1. The global parameters are not (yet) included in the output, but it would be easy to include them. Or we could add a separate feature to serialize the parameters to their own file.
1. The `Duration` type's default JSON serialization is a little odd. It has the advantage of having a corresponding `Deserialize` method. We can change it if we want.
1. The schema is up for grabs. Right now it's the JSON equivalent of this struct:

```rust
#[derive(Serialize)]
struct ProfilingData {
    date_time: SystemTime,
    execution_statistics: ExecutionStatistics,
    named_counts_headers: Vec<String>,
    named_counts_data: Vec<(String, usize, f64)>,
    named_spans_headers: Vec<String>,
    named_spans_data: Vec<(String, usize, Duration, f64)>,
}
```

In this PR, forecasting efficiency is hard coded. There is a separate PR #252 for the ability to add "custom" computed stats like forecasting efficiency.


Cargo.toml
```diff
@@ -3,11 +3,18 @@ name = "epi-isolation"
...
 
+[features]
+default = ["profiling"]
+
+profiling = ["humantime"]
+
 [dependencies]
-ixa = "0.3.0"
+#ixa = "0.3.0"
+ixa = { path = "../ixa" }
```
RobertJacobsonCDC, line 13: This PR uses the feature branch `RobertJacobsonCDC_488_pub_exec_stats`.

input/input.json
```diff
@@ -43,6 +43,7 @@
         {"min": 65,"probability": 0.05}
       ],
       "hospital_incidence_report_name": "hospital_incidence_report.csv"
-    }
+    },
+    "profiling_data_path": "profiling_data.json"
```
KOVALW, line 47: Should we add this default file name to the gitignore since JSONs are tracked? We're not doing anything with the created file just yet and individual level experiment outputs will be handled by the user later

src/profiling/file.rs
```diff
@@ -0,0 +1,62 @@
...
+}
+
+#[cfg(feature = "profiling")]
+pub fn write_profiling_data_to_file<P: AsRef<Path>>(
+    file_path: P,
+    execution_statistics: ExecutionStatistics,
+) -> std::io::Result<()> {
+    let container = profiling_data();
+    let profiling_data = ProfilingData {
+        date_time: SystemTime::now(),
```
KOVALW, line 34: It might be helpful to have this in a human-readable format instead of the default

KOVALW (commented):
Overall LGTM! The new code is straightforward. I do worry about the possible proliferation of tracked files because of the JSON files that get generated in the profiling data path. I think we should take some steps to prevent those from sneaking into PRs too frequently. Maybe by adding the default input file to the gitignore is sufficient? This wouldn't catch any time someone uses an output directory or prefix. Do we simply put the onus of avoiding this on the reviewer to make sure outputs don't enter PRs?

input/input.json
RobertJacobsonCDC: Added a blanket ignore pattern for the output directory based on our conversation.

KOVALW (approved):
LGTM!

PR #246, opened 2025-08-11
Gce bugfix infectiousness

confunguido:


KOVALW (commented):
This is a step in the right direction but I want to be careful about outright not listening to itinerary modifications. 

I think this is an alternative that works because we don't have people modifying itineraries beyond restriction or exclusion. If someone swaps the id of setting type, this has potential to break. 

I'm also unclear on how the rejection sampling for the increased infection attempt rate is being handled, people who are members with ratio 0.0 are not being filtered out of the infection attempt process.

PR #247, opened 2025-08-12
Preliminary report

RobertJacobsonCDC:
I can take a minute off (~20%) of the running time for 10M on my machine if we keep the original data types in the hash maps instead of converting to string vectors. It also takes 0.6 GiB less memory. Of course you have to use a different method for the initialization of the report, because the tabulator works on string vectors for some reason.

We are basically reproducing what the tabulator would do if we used it on the single derived property `PersonReportProperties`, so I need to think a little more about this to figure out the "right" way to implement it in the long run.

```
git:[RobertJacobsonCDC_gce_counts_report]
cargo run --release -- --config ./input/input_state_robert.json -f -o ./output --log-level epi_isolation=info
    Finished `release` profile [optimized] target(s) in 0.17s
     Running `target/release/epi-isolation --config ./input/input_state_robert.json -f -o ./output --log-level epi_isolation=info`
Loading global properties from: "./input/input_state_robert.json"
Logging enabled for epi_isolation at level INFO
━━━━ Execution Summary ━━━━
Max memory usage:        3.6 GiB
CPU time:                4m 8s 45ms
Wall time:               4m 24s 153ms 917us 42ns
Population:              10000022
Memory per person:       389 B
CPU time per person:     24us 804ns
Wall time per person:    26us 415ns


git:[gce_counts_report]
cargo run --release -- --config ./input/input_state_robert.json -f -o ./output --log-level epi_isolation=info
   Compiling epi-isolation v0.1.0 (/Users/rjacobson/Code/ixa-epi-isolation)
    Finished `release` profile [optimized] target(s) in 16.16s
     Running `target/release/epi-isolation --config ./input/input_state_robert.json -f -o ./output --log-level epi_isolation=info`
Loading global properties from: "./input/input_state_robert.json"
Logging enabled for epi_isolation at level INFO
━━━━ Execution Summary ━━━━
Max memory usage:        4.2 GiB
CPU time:                5m 28s 532ms
Wall time:               5m 47s 860ms 616us 625ns
Population:              10000022
Memory per person:       451 B
CPU time per person:     32us 853ns
Wall time per person:    34us 786ns
```


RobertJacobsonCDC:
In this version, I used a tuple instead of nested hash maps. Contrary to intuition, nested hash maps require _more_ memory, not less, because the data is very small relative to hash map overhead in this case, and you have to pay the cost of multiple hash map lookups. I also preserve the original types.

I wasn't sure if you were trying to guarantee zero counts appear in the report, but I didn't do that in my version. It's easy enough to restore that property of the code without any significant changes. Let me know if you want me to do it in this branch.

confunguido (approved):
This looks great! Thanks. Only thing is to figure out whether we want the 0s or not.

confunguido (approved):

PR #249, opened 2025-08-12
Allowing different symptom categories for constant rate function

erik-rosenstrom:
This PR enables for symptom categories to be randomly selected for individual symptom progressions for a constant infectiousness rate function.

This subsequently led to an update for testing the updated isolation guidance.


src/symptom_progression.rs
```diff
@@ -244,9 +246,16 @@ pub fn init(context: &mut Context) -> Result<(), IxaError> {
 
     // For isolation guidance, each rate function has a corresponding symptom improvement time
     // distribution, so we enforce a 1:1 relationship between the two.
-    context.register_parameter_id_assigner(Symptoms, |context, person_id| {
-        context.get_parameter_id(RateFn, person_id)
-    })?;
+    if RateFn.library_size(context) == 1 {
```
KOVALW, line 249: I would be more explicit and check that the global parameter `infectiousness_rate_fn` is a `RateFnType::Constant` here instead of using the `RateFn.library_size`.

erik-rosenstrom:
I think ideally, we should be able to switch between the empirical from file for symptoms progressions and an a way to specify a presymptomatic distribution and a symptom distribution. Basically a corollary to constant version of the infectiousness curve. I think that would give us the ability understand the effects of using symptom improvement times from the study.

KOVALW (approved):
LGTM!

PR #260, opened 2025-08-28
Connecting itinerary modifications to person properties

erik-rosenstrom:
This draft follows the example of transmission modifiers to connect modified itineraries to changes in person properties. It will also address multiple itinerary modifiers and have hospitalizations modify itineraries.

### DIFF
--- a/input/input.json
+++ b/input/input.json
@@ -31,17 +31,17 @@
       "facemask_efficacy": 0.2
     },
     "hospitalization_parameters": {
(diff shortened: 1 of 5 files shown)


KOVALW:
It would be helpful to have a team-level design meeting separate from the conversation on this PR (with motivating examples).

This PR could be split into settings refactor, itinerary manager, and policy manager (making sure that the last two have ownership over interventions such as school closures)

PR #263, opened 2025-09-03
Reports overhaul

KOVALW:
Refactors how reports are configured and generated in the codebase. Instead of specifying individual report file names and periods directly in parameters, reports are now defined as a structured list in the `input/input.json` and passed as a vector of `ReportType` enums. The initialization and validation of reports are centralized in the new `reports` module, which also separates out an incidence report that was previously combined with the prevalence `person_property_count` report. 

**Changes**

* Replaced individual report file name and period parameters in `input/input.json` and `Params` with a `reports` vector, using the new `ReportType` enum to specify report types, file names, and periods. 
* Removed obsolete fields and validation logic for individual report names and periods from `Params`, `HospitalizationParameters`, and related input validation.
* Separated out a new `incidence_report` module to generate periodic incidence reports, tracking infection status changes, symptom onset, and hospitalizations by age as before.
* Updated the test suite to verify correct report generation and configuration using the new system.


erik-rosenstrom (approved):
Looks good to me! One thought for the future I had would be to try and have tabulators specifiable from the input file. This could help speed up post processing of simulation results and the prevent use having to change the report implementation in the future if the calibration targets change.

PR #264, opened 2025-09-04
feat: Clarify intended computed statistics workflow.

RobertJacobsonCDC:
The main addition in this PR is a new top-level `computed_statistics` module where any custom computed statistics can be defined and initialized. It contains an `init()` function that is called from `main()` in the standard way.

This PR is intended to clarify the intended workflow for adding custom computed statistics. We can also add any tweaks we think we need to this PR as we review it.

- Moved forecasting efficiency statistic to a new module external to the profiling module to demonstrate intended workflow.
- Renamed `ProfilingDataContainer` to `ProfilingData` to clarify that it is not a data plugin.
- A couple of superficial lint fixes.

### DIFF
--- a/src/computed_statistics.rs
+++ b/src/computed_statistics.rs
@@ -43,9 +43,22 @@
 
 use crate::profiling::{
     add_computed_statistic, CustomStatisticComputer, CustomStatisticPrinter, ProfilingData,
-    ACCEPTED_INFECTION_LABEL, FORECASTED_INFECTION_LABEL,
 };
 
+/// The name of the distinguished accepted infection label. You don't need to make a constant
+/// for this, but it can keep you from introducing a bug because of a typo in the label.
+/// You would use the constant everywhere in your code instead of the string literal.
+pub const ACCEPTED_INFECTION_LABEL: &str = "accepted infection attempt";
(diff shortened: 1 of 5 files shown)


KOVALW (approved):
This is a lot clearer! I was able to quickly add a new computer and printer as a test using the percent of infected individuals that recover. I would maybe call the `src/computed_statistics.rs` file something like "forecast_efficiency_profiling" to be as straightforward as possible.

RobertJacobsonCDC:
I added a commit that moves the `ACCEPTED_INFECTION_LABEL`/`FORECASTED_INFECTION_LABEL` constants to the external `computed_statistics` module, since it doesn't make sense to have it in `profiling` anymore. Also, I replace the string literal usages of those two labels with their corresponding label constant.

Using a `pub const LABEL_CONSTANT: &'static str = "some label";` is completely optional, but I thought I would demonstrate it. I do recommend it though, because it is very easy to make a typo in the string literal that isn't caught by the compiler, whereas a typo in the constant name will be flagged. In fact, I caught a bug in one of the tests when I made the change for this PR! But for labels that are, for example, only used in a single place, it's not super important.

PR #272, opened 2025-09-15
Hospital age groups derived person property

erik-rosenstrom:
Given the change in https://github.com/CDCgov/ixa/pull/510 it is now possible to have hospital age groups be a derived person property, this means we can remove the hospital data container which stored the age group mapping.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ default = ["profiling"]
 profiling = ["humantime"]
 
 [dependencies]
-ixa = "0.3.2"
+ixa = "0.3.3"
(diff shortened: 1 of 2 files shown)


KOVALW (changes requested):
This is a big readability improvement. I'm getting comparable performance for runtime and memory usage. We need to think about better structures for `AgeGroups` and how to connect these across reports, hospitalization, and other risk categories, though.

The only change I'm requesting is that you do bump ixa in the Cargo.toml to 0.3.3 and add that here

KOVALW (approved):
LGTM

PR #275, opened 2025-09-17
Remove `reports` vector

KOVALW:
Switched from a vector of report enums to explicit, structured report parameters for each report type. 

**Changes**
* Replaces the `reports` vector of `ReportType` enums in `Params` with explicit fields for `prevalence_report`, `incidence_report`, and `transmission_report`, each using the new `ReportParams` struct. 
* Updates the JSON input format and test fixtures to use the new explicit report fields instead of the previous array-of-enums format. 
* Introduces the `ReportParams` struct, which includes `write`, `name`, and `period` fields, replacing the previous enum-based approach. Users can now use `"prevalence_report": {"write": false}` to drop the prevalence report, or keep the filename and period present while also setting to `false` to drop the report.
* Adds helper functions (`get_report_name`, `get_period_report_name`) and refactors the `init` function to validate report configuration, ensuring required fields are present and valid (e.g., `name` must be provided if `write` is true, `period` must be positive for periodic reports).
* Refactors all affected tests to use the new `ReportParams` structure and explicit report fields, ensuring test coverage for the new validation logic and configuration format.

### DIFF
--- a/input/input.json
+++ b/input/input.json
@@ -20,17 +20,17 @@
     "synth_population_file": "input/people_test.csv",
     "prevalence_report": {
         "write": true,
-        "name": "person_property_count.csv",
+        "filename": "person_property_count.csv",
         "period": 1.0
     },
     "incidence_report": {
         "write": true,
-        "name": "incidence_report.csv",
+        "filename": "incidence_report.csv",
         "period": 1.0
     },
     "transmission_report": {
         "write": true,
-        "name": "transmission_report.csv"
+        "filename": "transmission_report.csv"
     },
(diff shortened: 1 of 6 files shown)


input/input.json
```diff
@@ -18,19 +18,20 @@
...
-    ],
+    },
+    "incidence_report": {
+        "write": true,
+        "name": "incidence_report.csv",
+        "period": 1.0
+    },
+    "transmission_report": {
+        "write": true,
+        "name": "transmission_report.csv"
```
confunguido, line 33: Should we be more specific here, like filename?

confunguido (approved):
I left one comment but good to merge once addressed

PR #277, opened 2025-09-19
feat: add optional reporting delay to reports 

erik-rosenstrom:
This PR implements report specific reporting delays and is included in `ReportParams`. The reporting delay modifies the `t` value for each entry in the report.  This parameter is needed to model the epidemic delay when attempting to calibrate a guidance policy to observed hospital incidence data.

### Changes
- `reporting_delay` added as element to `ReportParams` struct. 
- getter functions updated to include `reporting_delay` which is then used in report initialization
- reporting_delay added to the `t` value 
- A prevalence test was updated to include a non zero reporting delay.

### Assumptions
- the reporting delay can be positive or negative
- The reporting delay is optional and defaults to 0
- Any required rounding of this value from an integer to a float will occur during post processing.
- Reports can have different reporting delays.

### DIFF
--- a/src/reports/mod.rs
+++ b/src/reports/mod.rs
@@ -31,7 +31,7 @@ fn get_report_name(params: &ReportParams) -> Result<Option<&str>, IxaError> {
     Ok(None)
 }
 
-fn get_period_delay_report_name(
+fn get_report_name_period_delay(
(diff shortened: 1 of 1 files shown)


confunguido:
We will implement this externally for now

PR #281, opened 2025-10-01
Define communities by census tract rather than PUMA

eqmooring:
## Background
Previously, the script `create_synthetic_population.R` used Public Use Microdata Areas (PUMAs) rather than census tracts as the geographic unit for community (even though the `input.json` file refers to `CensusTract` as a setting). Because PUMAs are much bigger than census tracts, one consequence of this approach is that the number of synthetic people per nominal "census tract" is potentially far larger than we see in other, more realistic synthetic population.

## What's changed
The script `create_synthetic_population.R` now assigns each household to a randomly selected census tract that in the real world is nested within that household's PUMA. (Census tracts are the building blocks of PUMAs, so a tract is never split between PUMAs.) The script performs simple random sampling without regard to the real-world number of households per census tract or any other known characteristics of the census tract.

A related change is that school and workplace locations are randomly sampled from census tracts in the state rather than the list of PUMAs. (As before, schools and workplaces are randomly assigned without regard to a synthetic person's PUMA or census tract.)

The structure and interpretation of the output files (e.g., `synth_pop_people_WY_10000,csv` has not changed. The key difference is that the first 11 digits of the columns `homeId`, `schoolId`, and `workplaceId` now refer to the GEOID of a real-world census tract (specifically, a concatenation of a 2-digit state code + 3-digit county code + 6-digit census tract code).

Another minor change (unrelated to PUMAs) is to adjust the default number of schools per person to a more realistic value.

## Outstanding issues

-  One consequence of this approach is that there is less variability in simulated households per census tract than observed in the real world. Real-world variability in census tract size can arise for various reasons, including the fact that some counties are very small and census tracts must be nested within counties. It would be feasible to obtain data on the real-world number of households per census tract and then use those numbers as weights when randomly selecting a census tract for a household, so that the simulated number of households per census tract mirrors more closely the real-world distribution of households per census tract. Alternatively, we could come up with a way to assign households to census tracts that more deterministically followed the distribution of households per census tract. It's unclear that these are necessary features, though, for the purpose of this basic synthetic population.
- The number of households per census tract goes down when the synthetic population is smaller and goes up when it's larger. (It's possible that the number of census tracts represented in the synthetic population can go down, but that's only likely to happen when the synthetic population size is much smaller than the corresponding real-world population size; by chance alone, most census tracts will get randomly selected at least once if there are far more households than census tracts.) Notably, this behavior is different from how schools and workplaces are generated, where the number is relative to the population size, We could create an approach that maintains the average number of households per census tract regardless of overall population size, but this would require ignoring some census tracts when the synthetic population size is small and, more problematically, inventing new census tracts that don't exist in the real-world when the simulated population size is larger. It's unclear if we want to go this direction.

## Resources
Here are a few resources that provide helpful background on census geographies.
https://www.census.gov/programs-surveys/geography/guidance/geo-identifiers.html
https://www.census.gov/programs-surveys/geography/guidance/geo-areas/pumas.html
https://www.census.gov/library/reference/code-lists/ansi.html


eqmooring:
@confunguido, I generated a synthetic population of 10,000,000 (still using WY as the underlying data) in 8 minutes and 15 seconds.

eqmooring:
I was running into difficulties with the command
```
tracts_st <- tracts(state = state_synth, year = year_synth)
```
for California (and possibly Texas, though I wasn't able to replicate that aspect).

Upon reviewing the documentation for this function (https://www.rdocumentation.org/packages/tigris/versions/2.2.1/topics/tracts), I suggest we want to use the argument `cb = TRUE`, which triggers downloading versions of the spatial data that are "cartographic boundaries" that are intended for plotting at 1:500,000 resolution. These are significantly smaller downloads, and this works for the state of California. The resulting dataframe, however, does not have the variables `INTPTLAT` and `INTPTLON`, so we have to calculate those as the centroids of the shapes.

Side note: It's unclear to me why latitude and longitude are even pulled when generating schools and workplaces; this doesn't seem to affect the output at all.

For illustration, here are two plots of the census tracts of Cheyenne, Wyoming and the surrounding area.
The first is with the generalized cartographic boundaries and the second is with the "exact" boundaries. I think for our purposes the differences are subtle and not important.

image
image

eqmooring:
I tried running with California as the state and a synthetic population of 40 million. The process was killed after about 50 minutes I think because it ran out of memory (7+ GB of RAM).

PR #286, opened 2025-10-28
Update setup

dinacmistry:
On a fresh install of the repo, both `poetry install` and `pre-commit` were failing for me and this seems to be systemic since Guido ran into the same. Proposing these changes to address some common issues that prevent a new contributor or user from onboarding smoothly.

After troubleshooting with Guido and Will, we figured out that running `pre-commit autoupdate` updates the pre-commit config file and allows `pre-commit run --all-files` to pass. 

For poetry, I think we just needed to bump the version of `pyarrow` and we shouldn't need to specify the branch of `abmwrappers` that uses uv for dependency management.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,7 +2,7 @@ repos:
 #####
 # Basic file cleanliness
 -   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v5.0.0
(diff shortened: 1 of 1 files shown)


KOVALW (approved):
LGTM - I was able to install the repo from scratch, run the main example and run the replicate.py example

PR #290, opened 2025-11-20
Running scenarios with a calibrated model

erik-rosenstrom:
This PR builds on #288 to enable the running of griddle defined scenarios using a calibrated model. 

## What's changed

- `scenarios.py` uses abmwrapper's method to generate scenario subexperiments where each subexperiment corresponds to a replicate of a scenario. Scenarios are run locally in serial. 
- The pdf output associated with calibration has a new section showing incidence and cumulative differences across the scenarios. Figure examples are shown in the comments below

## Further Questions

- Automate pickle transition when moving from calibration to scenarios

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -3,8 +3,8 @@ use crate::parameters::{
 };
 use indexmap::set::IndexSet;
 use ixa::{
-    define_data_plugin, define_rng, trace, Context, ContextPeopleExt, ContextRandomExt, HashMap,
-    HashMapExt, HashSet, HashSetExt, IxaError, PersonId, PluginContext,
+    define_data_plugin, define_rng, profiling::open_span, trace, Context, ContextPeopleExt,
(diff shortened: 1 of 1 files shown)


experiments/simple-fits/wyoming-incidence/docs/output.qmd
```diff
@@ -353,6 +353,186 @@ plt.subplots_adjust(bottom=0.1, left=0.1, top=0.7, right=0.7, wspace=0.2, hspace
 plt.show()
 
 ```
+
+# Scenario Analysis
+
+A scenario analysis is conducted that studies effects of the update and previous isolation guidance on hospitalizations. The guidances are compared to a counterfactual "No Guidance", which is the context for the model calibration. Two relavent policy parameters are studied in set of scenarios that are enumerated in the following table.
```
KOVALW:
```suggestion
The scenario analysis assesses the effects of the update and previous isolation guidance on hospitalizations in the calibrated "No Guidance" model. The table below shows the counterfactual scenarios and the relevant parameters varied by scenario.
```

PR #301, opened 2026-01-26
typographic and other minor revisions to documentation

eqmooring:
## Background
I was reading through the documentation for this repo (as part of preparing to work on integration tests) and I noticed a few typographic and other minor things, so I figured I would propose revisions as I went.

## What's changed
I have made edits to many of the documentation `.md` files, but I have not created or removed any of them. Most of the edits are typographic in nature, but a few are a bit more substantive.

## Outstanding issues
It might be good to note where in the inputs / code for this is specified in the code base (also, in the individual-level analysis, only category 1 was "moderate"--because those are the persons with shortness of breath):
https://github.com/CDCgov/ixa-epi-isolation/blob/0db097f56550cc49153c9d7f0d43f3a483c16547/docs/intervention-policies.md?plain=1#L51


erik-rosenstrom (approved):
LGTM. This is a typo in the documentation. In line 308 of the `symptom_progression.rs` the `severe` symptom progression property is defined. It is defined only as category 1. This is what is used in the intervention policies.

